PR #9, opened 2023-08-07
feat: deployment health server

### LINKED ISSUE #2: feat: Deployment health server
Deployment health server at `subgraphs/health/:deployment` to 
  - [ ] query graph node status endpoint for deployment's indexing status
  - [ ] process graph node's block number results for formatted response

hopeyen:
resolves #2

### DIFF
--- a/service/src/server/routes/deployment.rs
+++ b/service/src/server/routes/deployment.rs
@@ -1,7 +1,5 @@
 use axum::{http::StatusCode, response::IntoResponse, Extension, Json};
 
-use hyper::http::HeaderName;
-
 use serde::{Deserialize, Serialize};
 use serde_json::json;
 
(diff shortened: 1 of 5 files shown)


service/src/server/routes/deployment.rs
```diff
@@ -0,0 +1,149 @@
...
+                internal_server_error_response("Unknown error")
+            }
+        }
+        Err(e) => internal_server_error_response(&e.to_string()),
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct IndexingStatus {
+    health: String,
```
aasseman, line 87: Could it make sense to go the extra mile and use an Enum here?

```rs
#[derive(Debug, Clone, Serialize, Deserialize)]
#[allow(non_camel_case_types)] // Need exact field names to match with GQL response
enum SubgraphHealth {
    healthy,
    unhealthy,
    failed
}
```

service/src/server/routes/mod.rs
```diff
@@ -37,3 +40,13 @@ pub fn bad_request_response(error_body: &str) -> Response {
     )
         .into_response()
 }
+
+/// Create response for an internal server error
+pub fn internal_server_error_response(error_body: &str) -> Response {
+    (
+        StatusCode::INTERNAL_SERVER_ERROR,
+        axum::response::AppendHeaders([(HeaderName::from_static("graph-attestable"), "false")]),
```
aasseman, line 48: I don't think that header makes sense here

aasseman (changes requested):

service/src/server/routes/deployment.rs
```diff
@@ -0,0 +1,149 @@
...
+                }
+
+                if let Ok((latest, head)) = block_numbers(status) {
+                    if latest > head - 5 {
+                        (
+                            StatusCode::OK,
+                            axum::response::AppendHeaders([(
+                                HeaderName::from_static("graph-attestable"),
+                                "true",
+                            )]),
```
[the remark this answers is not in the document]
hopeyen: ah yessss, will remove the graph-attestable headers

service/src/server/routes/deployment.rs
hopeyen: another idea is maybe we could use [serde alias](https://serde.rs/field-attrs.html), but I'm fine with the current way:)

PR #21, opened 2023-08-08
fix: subgraph queries responses

aasseman:
Mostly to make it behave like the original `indexer-service`.

### DIFF
--- a/service/src/query_processor.rs
+++ b/service/src/query_processor.rs
@@ -89,16 +89,16 @@ pub struct Signature {
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[allow(non_snake_case)] // Need exact field names for JSON response
 pub struct QueryResult {
-    pub graphQLResponse: String,
+    #[serde(rename = "graphQLResponse")]
+    pub graphql_response: String,
     pub attestation: Option<Signature>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
-#[allow(non_snake_case)] // Need exact field names for JSON response
 pub struct UnattestedQueryResult {
-    pub graphQLResponse: String,
+    #[serde(rename = "graphQLResponse")]
+    pub graphql_response: String,
     pub attestable: bool,
 }
 
--- a/service/src/query_processor.rs
+++ b/service/src/query_processor.rs
@@ -114,7 +114,6 @@ pub struct Response<T> {
 #[derive(Debug)]
 pub struct FreeQuery {
     pub subgraph_deployment_id: SubgraphDeploymentID,
-    // TODO: Use Arc<String> instead of String?
     pub query: String,
 }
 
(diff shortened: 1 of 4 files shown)


service/src/query_processor.rs
```diff
@@ -71,20 +71,25 @@ impl ToString for SubgraphDeploymentID {
...
+    pub graphQLResponse: String,
+    pub attestation: Option<Signature>,
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[allow(non_snake_case)] // Need exact field names for JSON response
 pub struct UnattestedQueryResult {
-    pub graphql_response: String,
+    pub graphQLResponse: String,
```
hopeyen, line 92: Perhaps we can use serde alias here to match JSON response?

service/src/query_processor.rs
```diff
@@ -100,6 +105,7 @@ pub struct Response<T> {
 #[derive(Debug)]
 pub struct FreeQuery {
     pub subgraph_deployment_id: SubgraphDeploymentID,
+    // TODO: Use Arc<String> instead of String?
```
hopeyen, line 108: maybe worthwhile, especially if you think the query string is shared across threads or parts of the program, or if it is used multiple times, or should be long living

service/src/query_processor.rs
aasseman: Oh cool didn't know about that solution. I know `strum` can do that too, I used it in the TAP Aggregator service.

service/src/query_processor.rs
aasseman: Yup, in particular I noticed that
https://github.com/hopeyen/indexer-service-rs/blob/c604c7a258fad37ef74abd20823571d5f4db9e3e/service/src/graph_node.rs#L28-L32
needs the cloning.

aasseman:
Rebased onto latest main

service/src/query_processor.rs
aasseman: Fix in [0c02b88](https://github.com/graphops/indexer-service-rs/pull/21/commits/0c02b88971474da1147ffd1ca8565fefbe2c9378)

hopeyen (approved):
🐈‍⬛  (might need another rebase to main:p

service/src/query_processor.rs
aasseman: I played around, went down the rabbit hole, and gave up: [8b9b531](https://github.com/graphops/indexer-service-rs/pull/21/commits/8b9b531ea6b59ef34563dc18ca4f9590727890a1)

hopeyen: haha we could leave the TODO comment in for a future refactor potentially?

aasseman: Well, I'm not even sure that Arc is the solution here. So I'm of the opinion of sticking to the YAGNI rule, we can do something about it when we'll actually need it.

hopeyen (approved):
🚢

PR #27, opened 2023-08-12
feat: network_query w/ gql and vars

aasseman:
To make it easier to send out network queries for things like allocation tracking, etc.
Also made the API slightly more explicit.


hopeyen (approved):
good idea to have a raw inner for the queries!

PR #28, opened 2023-08-12
refactor: move graph-node instance to main

aasseman:
Small change that will make it easier to share the graph-node instance wrapper with other flows, such as the allocation monitor.


hopeyen (commented):
The refactoring of graph-node made sense, though I think Graph node / Query processor shouldn't track the endpoint for network subgraph. Maybe it is worth having a separate network_subgraph client to execute queries with?

aasseman:
Rebased onto latest main

PR #30, opened 2023-08-15
Fix graph node instance

aasseman:
Details in commit messages

### DIFF
--- a/service/src/graph_node.rs
+++ b/service/src/graph_node.rs
@@ -30,7 +30,7 @@ struct GraphQLQuery {
 impl GraphNodeInstance {
     pub fn new(endpoint: &str, network_subgraph_id: &str) -> GraphNodeInstance {
         let subgraphs_base_url = Url::parse(endpoint)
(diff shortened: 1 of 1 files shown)


hopeyen:
The code looks good, would you be able to post some examples of what the expected requests look like?
I'm asking because the previous format doesn't seem to process anymore and I started to get "Empty reply from server"

aasseman:
:red_circle: as expected!

aasseman:
And now :green_circle: !
Though it looks that Semantic PR is not set to accept revert commits

hopeyen (approved):
lgtm!
I will request an update to our org semantics settings for revert commits, you could also rewrite the commit message, or drop the commit (don't know if it is more conventional to

aasseman:
> lgtm! I will request an update to our org semantics settings for revert commits, you could also rewrite the commit message, or drop the commit (don't know if it is more conventional to use Merge and Revert commits, I usually prefer rebase and squash commits for a linear and concise git history

Yup, we can definitely squash this one

PR #36, opened 2023-08-18
ci: improve rust cache

aasseman:
Experiments here: https://github.com/aasseman/indexer-service-rs/pull/1

tl;dr reduced the time by more than an order of magnitude


hopeyen (approved):
nice🙌!!!

PR #37, opened 2023-08-22
refactor: splits network subgraphs query to NetworkSubgraph

hopeyen:
Isolate network subgraph from query processor, so that network queries are separated from subgraph queries.
Network subgraph serves free queries with auth token, whereas paid queries should come through `subgraphs/id/:id` with the deployment ID and receipts

### DIFF
--- a/service/src/common/network_subgraph.rs
+++ b/service/src/common/network_subgraph.rs
@@ -1,0 +1,195 @@
+// Copyright 2023-, GraphOps and Semiotic Labs.
+// SPDX-License-Identifier: Apache-2.0
+
+use std::sync::Arc;
+
+use reqwest::{header, Client, Url};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+
+use crate::query_processor::{QueryError, UnattestedQueryResult};
+
+#[derive(Debug, Serialize, Deserialize, PartialEq)]
+pub struct Response<T> {
+    pub result: T,
+    pub status: i64,
+}
+
+/// Network subgraph query wrapper
+///
+/// This is Arc internally, so it can be cloned and shared between threads.
+#[derive(Debug, Clone)]
+pub struct NetworkSubgraph {
+    client: Client, // it is Arc
+    network_subgraph_url: Arc<Url>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+struct GraphQLQuery {
+    pub query: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub variables: Option<Value>,
+}
+
+impl NetworkSubgraph {
+    pub fn new(
+        graph_node_query_endpoint: Option<&str>,
+        deployment: Option<&str>,
+        network_subgraph_url: &str,
+    ) -> NetworkSubgraph {
+        //TODO: Check indexing status of the local network subgraph deployment
+        //if the deployment is healthy and synced, use local_network_subgraph_endpoint
+        let _local_network_subgraph_endpoint = match (graph_node_query_endpoint, deployment) {
+            (Some(endpoint), Some(id)) => {
+                Some(NetworkSubgraph::local_deployment_endpoint(endpoint, id))
+            }
+            _ => None,
+        };
+
+        let network_subgraph_url =
+            Url::parse(network_subgraph_url).expect("Could not parse network subgraph url");
+
+        let client = reqwest::Client::builder()
+            .user_agent("indexer-service")
+            .build()
+            .expect("Could not build a client to graph node query endpoint");
+
+        NetworkSubgraph {
+            client,
+            network_subgraph_url: Arc::new(network_subgraph_url),
+        }
+    }
+
+    pub fn local_deployment_endpoint(graph_node_query_endpoint: &str, deployment: &str) -> Url {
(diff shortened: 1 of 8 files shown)


service/src/common/network_subgraph.rs
```diff
@@ -0,0 +1,192 @@
...
+            .build()
+            .expect("Could not build a client to graph node query endpoint");
+
+        NetworkSubgraph {
+            client,
+            network_subgraph_url: Arc::new(
+                network_subgraph_url.expect("Could not query network subgraph"),
+            ),
+        }
+    }
```
aasseman, line 65: I have a hard time understanding this piece of code. I'd imagine that the arguments should be only `graph_node_query_endpoint` and `deployment` (both non optional) *or* only `network_subgraph_url` (non optional).

service/src/common/network_subgraph.rs
```diff
@@ -0,0 +1,192 @@
...
+        NetworkSubgraph::new(
+            Some(GRAPH_NODE_STATUS_ENDPOINT),
+            Some(NETWORK_SUBGRAPH_ID),
+            Some(NETWORK_SUBGRAPH_URL),
+        )
+    }
+
+    #[tokio::test]
+    #[ignore] // Run only if explicitly specified
+    async fn test_network_query_local() {
```
aasseman, line 156: This is not local, as it is querying thegraph.com. So perhaps this one should be `test_network_query` and the other one `test_network_query_mock`?

service/src/config.rs
```diff
@@ -185,10 +185,9 @@ pub struct NetworkSubgraph {
         long,
         value_name = "network-subgraph-endpoint",
         env = "NETWORK_SUBGRAPH_ENDPOINT",
-        default_value_t = String::from("https://api.thegraph.com/subgraphs/name/graphprotocol/graph-network-goerli"),
         help = "Endpoint to query the network subgraph from"
     )]
-    pub network_subgraph_endpoint: String,
+    pub network_subgraph_endpoint: Option<String>,
```
aasseman, line 190: Why making it optional? I'd think that the indexer-service cannot operate without being able to query the network subgraph.

aasseman (changes requested):
At this point, it looks like `GraphNodeInstance` will only be used by the query processor. So we could consider removing `GraphNodeInstace` altogether and move its functionality inside query processor, perhaps.

service/src/common/network_subgraph.rs
```diff
@@ -0,0 +1,192 @@
+// Copyright 2023-, GraphOps and Semiotic Labs.
+// SPDX-License-Identifier: Apache-2.0
+
+use reqwest::{header, Client, Url};
+use serde::{Deserialize, Serialize};
+use serde_json::Value;
+use std::sync::Arc;
```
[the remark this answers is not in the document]
hopeyen: good to know! wishing `fmt` would take care of this

service/src/config.rs
hopeyen: ah yes you are right, changing it back!

service/src/common/network_subgraph.rs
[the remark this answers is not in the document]
aasseman: It does, but not by default: https://rust-lang.github.io/rustfmt/?version=v1.4.38&search=#group_imports
IIRC a few months ago it was marked as experimental/dangerous, but there is no warnings anymore.

service/src/common/network_subgraph.rs
hopeyen: My thinking around this is to track network subgraph at both the local deployment query endpoint and a remote endpoint. Although `graph_node_query_endpoint` and `deployment` should be used after we have indexing status resolver and be able to check that the deployment is synced. 
if all three inputs are provided then I think we could track both methods of querying network subgraph and use the remote endpoint as a backup when the local deployment isn't synced

aasseman: OK I see, I forgot about the backup remote endpoint bit.

service/src/common/network_subgraph.rs
```diff
@@ -0,0 +1,192 @@
...
+        graph_node_query_endpoint: Option<&str>,
+        deployment: Option<&str>,
+        network_subgraph_url: &str,
+    ) -> NetworkSubgraph {
+        //TODO: Check indexing status of the local network subgraph deployment
+        //if the deployment is healthy and synced, use local_network_subgraph_endpoint
+        let _local_network_subgraph_endpoint = match (graph_node_query_endpoint, deployment)
+        {
+            (Some(_endpoint), Some(id)) =>
+            Some(Url::parse(graph_node_query_endpoint.unwrap())
```
aasseman, line 45: I think you meant it to be like this:
```suggestion
            Some(Url::parse(_endpoint)
```

aasseman (changes requested):

service/src/common/network_subgraph.rs
hopeyen: yikes! thank u

aasseman (changes requested):
Needs fmt and tests fixed and we're good!

hopeyen:
rebased to main and fixed the tests:)!

aasseman (approved):
:shipit:

PR #38, opened 2023-08-23
feat: attestation signers added to allocation monitor

aasseman:
There is some weirdness due to type conversions to interface with the `native` crate. I think it'd be nice to refactor `native` to have it rely on external dependencies, to make it easier to understand and maintain in the future.


hopeyen (approved):
looks good! think allocation monitor is a great idea versus signers.

aasseman:
Rebased onto latest main

Issue #4, opened 2023-08-07
Feat: Record metrics


hopeyen:
- [x] register metrics, serve metrics port that can be scrapped by Prometheus
- [x] record queries metrics
- [ ] record cost model metrics
- [ ] record indexer errors
- [ ] Agreeing with [comments](https://github.com/graphops/indexer-service-rs/pull/39#issuecomment-1692225428) from @aasseman , we can further improve metrics with
  > query duration be discarded if there's an error
  > FAILED_QUERIES be incremented at all the points where the functions returns with an error

PR #39, opened 2023-08-24
Feat: record query metrics

hopeyen:
Track query volume and performance like the original indexer service.
Some metrics from the original are registered but not used so they are not included here

Part of https://github.com/graphops/indexer-service-rs/issues/4

For later: Add cost model metrics once there is cost server

### DIFF
--- a/service/src/metrics/mod.rs
+++ b/service/src/metrics/mod.rs
@@ -1,5 +1,6 @@
 // Copyright 2023-, GraphOps and Semiotic Labs.
 // SPDX-License-Identifier: Apache-2.0
+
 use autometrics::{encode_global_metrics, global_metrics_exporter};
 use axum::http::StatusCode;
 use axum::routing::get;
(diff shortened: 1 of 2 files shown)


aasseman:
:thinking: shouldn't the query duration be discarded if there's an error? I don't know what the choice was in the original indexer-service though.

aasseman:
And I also have similar question about `FAILED_QUERIES`. Shouldn't it be incremented at all the points where the functions returns with an error?

hopeyen:
The typescript version uses the syntax `try { start_timer and stuff } catch { error_handle } finally { stop_timer }` in which `finally` should happen regardless of `try` result, so it does record errored out queries. 
Does it make more sense to just discard? maybe I add another label with an Enum on the results🤔?

aasseman:
Yup I see, thanks for the pointers, yours indeed behaves like the original. IMO the original is wrong though :grimacing: .
Should we just go ahead with this, and perhaps track a potential rework of the query processing monitoring logic in the issues?

aasseman (approved):

hopeyen:
I agree with you 😂 We can change it up later, will add a note to the metrics tracking issue #4

PR #42, opened 2023-08-29
refactor: splitting off  `AttestationSigners`

### LINKED ISSUE #41: Consider splitting `AllocationMonitor` from `AttestationSigners`, etc
Would it make sense?
I'm considering this because TAP will also need to watch for new allocations, and having lots of TAP stuff inside of `AllocationMonitor` may be a little bit of an anti-pattern.
For such a change, I was looking into leveraging `tokio::sync::broadcast` or `tokio::sync::watch`, and the other components would just have to subscribe to know when the current set of eligible allocations has changed.

That may be overkill though :shrug:

aasseman:
Now can subscribe to `AllocationMonitor` updates, so `AttestationSigners` knows when to update its signers list. Makes way for TAP to also watch the `AllocationMonitor`.

Note: Moved some test vectors out of allocation_monitor.rs to share them amongst test files.

Fixes #41

### DIFF
--- a/service/src/allocation_monitor.rs
+++ b/service/src/allocation_monitor.rs
@@ -1,21 +1,18 @@
 // Copyright 2023-, GraphOps and Semiotic Labs.
 // SPDX-License-Identifier: Apache-2.0
 
-use std::{collections::HashMap, sync::Arc};
+use std::sync::Arc;
 
 use anyhow::Result;
-use ethereum_types::U256;
+
 use ethers::types::Address;
 
 use log::{info, warn};
-use native::attestation::AttestationSigner;
+
+use tokio::sync::watch::{Receiver, Sender};
 use tokio::sync::RwLock;
 
-use crate::{
-    common::allocation::{allocation_signer, Allocation},
-    common::network_subgraph::NetworkSubgraph,
-    util::create_attestation_signer,
-};
+use crate::{common::allocation::Allocation, common::network_subgraph::NetworkSubgraph};
 
 #[derive(Debug)]
 struct AllocationMonitorInner {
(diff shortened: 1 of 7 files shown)


service/src/allocation_monitor.rs
```diff
@@ -579,7 +369,8 @@ mod tests {
             std::env::var("NETWORK_SUBGRAPH_ID").expect("NETWORK_SUBGRAPH_ID not set");
         let indexer_address = std::env::var("INDEXER_ADDRESS").expect("INDEXER_ADDRESS not set");
         // If you don't specify the correct mnemonic, the attestation signers won't be found
-        let indexer_mnemonic = std::env::var("INDEXER_MNEMONIC").expect("INDEXER_MNEMONIC not set");
+        let _indexer_mnemonic =
```
hopeyen, line 372: Can this be removed?

service/src/allocation_monitor.rs
```diff
@@ -24,44 +21,42 @@ struct AllocationMonitorInner {
...
-    dispute_manager: Address,
+    watch_sender: Sender<()>,
+    watch_receiver: Receiver<()>,
 }
 
+#[cfg_attr(test, faux::create)]
 #[derive(Debug, Clone)]
 pub struct AllocationMonitor {
-    monitor_handle: Arc<tokio::task::JoinHandle<()>>,
+    _monitor_handle: Arc<tokio::task::JoinHandle<()>>,
```
hopeyen, line 31: what do we expect this to do? 

Is it right to think of `AllocationMonitorInner` as something that just keeps information including watcher and receiver while `AllocationMonitor` is doing the actual monitoring?

service/src/attestation_signers.rs
```diff
@@ -0,0 +1,162 @@
...
+    ) -> Self {
+        let inner = Arc::new(AttestationSignersInner {
+            attestation_signers: Arc::new(RwLock::new(HashMap::new())),
+            allocation_monitor,
+            indexer_mnemonic,
+            chain_id,
+            dispute_manager,
+        });
+
+        let _update_loop_handle = {
```
hopeyen, line 46: Similar question here on why do we keep _update_loop_handle for attestation signers?

service/src/allocation_monitor.rs
aasseman: Yes! Actually I don't remember adding the underscore there, weird.

aasseman: Fix in a410943128cabe8166880b041bb287714e770db0

service/src/allocation_monitor.rs
aasseman: The main reason I created an "inner" is to make this possible:
https://github.com/graphops/indexer-service-rs/blob/a410943128cabe8166880b041bb287714e770db0/service/src/allocation_monitor.rs#L58-L63

Basically I needed the spawned thread not depend on `AllocationMonitor`, because it is itself belonging to `AllocationMonitor`.
I could just clone all the values that the thread needs, but the "inner" makes it visually cleaner. Also makes it cleaner to Arc clone `AllocationMonitor`.

tl;dr it's mostly easthetics, I guess

aasseman: As for the handle, I don't like the idea of not explicitely keeping it somewhere. Is the thread even guaranteed to continue existing if the handle is lost? (I actually don't know yet what's going on at a lower level here)

hopeyen: And you are keeping track of the handle so it won't "roam free"? Would we need to specifically shutdown the spawned threads when we get termination or exit signal?

aasseman: > Would we need to specifically shutdown the spawned threads when we get termination or exit signal?

Well, that's a good question. I guess we'll see when we tie everything together :grin:

aasseman: Calling the Rust masters! Would @axiomatic-aardvark know?

axiomatic-aardvark: I think it makes sense either way, the thread will continue even without the handle, but it does seem more idiomatic to keep track of it. And as @hopeyen suggested that allows for a shutdown signal (maybe a `Arc<AtomicBool>`) to be used to break out of the loop

hopeyen (approved):
coolio! we can find out what happens when the components get all tied together

PR #43, opened 2023-08-31
indexer management client with db connection and cost server

hopeyen:
Resolves https://github.com/graphops/indexer-service-rs/issues/19 and https://github.com/graphops/indexer-service-rs/issues/3
some quick notes
- Simple indexer management client is now tracking both NetworkSubgraph and postgres connection
- Define cost model schema and resolvers with postgres and graphQL types, with global cost model fallback when specific deployments are queried
- No migration in indexer service; indexer agent is solely responsible for database management.


service/src/server/routes/cost.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use crate::{common::indexer_management_client::schema::CostSchema, server::ServerOptions};
+
+pub(crate) async fn graphql_handler(
+    req: GraphQLRequest,
+    Extension(schema): Extension<CostSchema>,
+    Extension(server_options): Extension<ServerOptions>,
+) -> GraphQLResponse {
+    let response = async move { schema.execute(req.into_inner().data(server_options)).await }.await;
+    response.into()
```
aasseman, line 15: That should be equivalent
```suggestion
    schema.execute(req.into_inner().data(server_options)).await.into()
```

aasseman, line 15: That should be equivalent
```suggestion
    schema.execute(req.into_inner().data(server_options)).await.into()
```

aasseman, line 15: That should be equivalent
```suggestion
    schema.execute(req.into_inner().data(server_options)).await.into()
```

service/src/common/indexer_management_client/resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+            },
+        ]
+    }
+
+    #[sqlx::test]
+    #[ignore]
+    async fn success_cost_models(pool: PgPool) {
+        _ = setup_cost_models_table(&pool).await;
+        let expected_models = simple_cost_models();
+        _ = add_cost_models(&pool, expected_models.clone()).await;
```
aasseman, line 217: These functions do not return a value. Applies to all the tests below too.
```suggestion
        setup_cost_models_table(&pool).await;
        let expected_models = simple_cost_models();
        add_cost_models(&pool, expected_models.clone()).await;
```

service/src/common/indexer_management_client/resolver.rs
```diff
@@ -0,0 +1,343 @@
...
+use sqlx::PgPool;
+
+use super::schema::CostModel;
+use crate::common::{
+    indexer_error::{IndexerError, IndexerErrorCause, IndexerErrorCode},
+    types::SubgraphDeploymentID,
+};
+
+/// Query postgres indexer management server's cost models
+/// Filter on deployments if it is not empty, otherwise return all cost models
```
aasseman, line 13: Needs a bit more explanation about what happens wrt. the global model and empty vs. non-empty `deployments`

aasseman (changes requested):
LGTM overall, just a few little things (see code comments).

Things that could be explored later:
- It'd be nice to have some tests on the graphql server, with request/response test vectors.
- Perhaps some migrations could be added only for the purposes of testing. I couldn't build on my machine (sqlx complaining) and had to run a PG instance and create the table myself. Though I'm confused why the CI didn't run into that problem.
- Related to the point above, we could set-up the CI such that there is a PG instance available, to avoid ignoring all the tests using sqlx.

hopeyen:
Thanks for the review @aasseman !
I've addressed the comments and regarding to things for later
> tests on the graphql server, with request/response test vectors.
Agreed and added an item on https://github.com/graphops/indexer-service-rs/issues/15

> Perhaps some migrations could be added only for the purposes of testing. I couldn't build on my machine (sqlx complaining) and had to run a PG instance and create the table myself. Though I'm confused why the CI didn't run into that problem.

Would it be because those tests had `[ignore]` and `cargo llvm-cov test` doesn't test the ignored? 
I've added a `migrations` folder for you to try out if you would like

> Related to the point above, we could set-up the CI such that there is a PG instance available, to avoid ignoring all the tests using sqlx.
Agreed as well, it would just need to run `sqlx migrate run` once the instance is available

aasseman (changes requested):
One last thing, perhaps adding a `migrations/README.md` that explains that the migrations are there only for testing?

hopeyen:
> adding a migrations/README.md that explains that the migrations are there only for testing

Good idea! added in https://github.com/graphops/indexer-service-rs/pull/43/commits/478182b4a7174819756197f277db9f0b2c82224c

aasseman (approved):
:shipit:

PR #52, opened 2023-09-18
Remove indexer management GraphQL API

Jannis:
This API is not supposed to be exposed by the indexer service. It is something that only indexer agent runs to give indexer CLI and other tools the ability to configure the indexer agent behavior.


Jannis:
@aasseman @hopeyen Fixed up the linter error, so this should be passing all checks now.

service/src/server/mod.rs
```diff
@@ -92,12 +88,6 @@ pub async fn create_server(
             get(routes::deployment::deployment_health
                 .layer(AddExtensionLayer::new(slow_ratelimiter()))),
         )
-        .route(
```
hopeyen, line 95: Should `indexer-service` not serve cost models at all?

hopeyen:
`indexer-service` doesn't serve cost models at all? Does the gateway request cost_models from this API?

service/src/server/routes/cost.rs
```diff
@@ -1,10 +1,43 @@
...
+        &self,
+        ctx: &Context<'_>,
+        deployment: String,
+    ) -> Result<Option<CostModel>, IndexerError> {
+        let pool = &ctx.data_unchecked::<ServerOptions>().indexer_management_db;
+        indexer_management::cost_model(pool, &deployment).await
+    }
+}
 
 pub(crate) async fn graphql_handler(
```
hopeyen, line 42: In the spirit of remove indexer management graphQL API, perhaps we remove `graphql_handler` and use a different handler for the `cost` requests?

Jannis: This is the public cost model API that indexer-service exposes as well — and that _is_ a GraphQL API. So this is correct.

hopeyen (approved):

PR #53, opened 2023-09-18
Refactor code into indexer-common crate

Jannis:
This moves out much of the common code into an `indexer-common` crate. The next step is likely to make more clear how one would go about implementing a different kind of indexer service.


aasseman:
This is awesome, I actually just barely started working on this to share components with the "`indexer-tap-agent`" (or better name if someone feels inspired)

Jannis:
@aasseman Ok, I expect this to pass the checks.

I didn't want to move the escrow monitor just yet. To avoid that, I had to split up the `test_vectors`. I think the TAP manager tests needed the allocation monitor and so I had to add a `mock` feature to the `indexer-common` crate to export the `faux` code generated for it across for testing. Probably something we can clean up again later.

aasseman:
> @aasseman Ok, I expect this to pass the checks.
> 
> I didn't want to move the escrow monitor just yet. To avoid that, I had to split up the `test_vectors`. I think the TAP manager tests needed the allocation monitor and so I had to add a `mock` feature to the `indexer-common` crate to export the `faux` code generated for it across for testing. Probably something we can clean up again later.

I agree on all points. The mocking was what slowed me down considerably in my attempt to split things off into a common crate this weekend. I can take a stab at the EscrowMonitor in particular in another PR. I'm the only one depending on this for now so you don't need to struggle with this.

common/src/attestations/signers.rs
```diff
@@ -21,7 +19,7 @@ pub struct AttestationSigners {
 }
 
 #[derive(Debug)]
-struct AttestationSignersInner {
+pub(crate) struct AttestationSignersInner {
```
aasseman, line 22: We don't need to make this public as of now.
```suggestion
struct AttestationSignersInner {
```

common/src/attestations/signers.rs
```diff
@@ -55,7 +53,7 @@ impl AttestationSigners {
         }
     }
 
-    async fn update_attestation_signers(inner: Arc<AttestationSignersInner>) {
+    pub(crate) async fn update_attestation_signers(inner: Arc<AttestationSignersInner>) {
```
aasseman, line 56: Same as above
```suggestion
    async fn update_attestation_signers(inner: Arc<AttestationSignersInner>) {
```

aasseman (changes requested):
LGTM except for 2 small things.
Also I think this should wait for #52 since this branch is based on top of it.

common/src/attestations/signers.rs
Jannis: @aasseman You were right. Fixed!

common/src/attestations/signers.rs
Jannis: Fixed in the latest commit.

Jannis:
Rebased after #52 was merged. Let's see if the tests still pass.

service/src/server/routes/network.rs
```diff
@@ -38,14 +39,25 @@ pub async fn network_queries(
...
-        _ => bad_request_response("Bad response from Graph node"),
+    if response.status().is_success() {
+        (
+            StatusCode::OK,
+            Json(
+                response
+                    .json::<Value>()
+                    .await
+                    // FIXME: Don't use expect here
+                    .expect("Failed to parse network subgraph query result"),
```
hopeyen, line 56: use another `bad_request_response`

aasseman (approved):

aasseman (approved):

PR #54, opened 2023-09-21
Refactor to use eventuals for monitoring network data

Jannis:
Eventuals are like `tokio::sync::watch`, except they only trigger an update downstream if their value changes, not every time a new value is written into it. It's a nice primitive to ensure we always have the latest value available and only do something when there are changes.

This switches to eventuals for the following things:
- Monitoring the indexer's own allocations. This outputs an update to date list of active and recently closed allocations and updates every 30s.
- Creating attestation signers. This takes the indexer's allocations and ensures that we can always sign attestations for queries made against the subgraphs that were allocated to.

### DIFF
--- a/common/src/allocations/mod.rs
+++ b/common/src/allocations/mod.rs
@@ -2,19 +2,14 @@
 // SPDX-License-Identifier: Apache-2.0
 
 use alloy_primitives::Address;
-use anyhow::Result;
-use ethers::signers::coins_bip39::English;
-use ethers::signers::{MnemonicBuilder, Signer, Wallet};
-use ethers_core::k256::ecdsa::SigningKey;
 use ethers_core::types::U256;
-use serde::Deserialize;
-use serde::Deserializer;
+use serde::{Deserialize, Deserializer};
 
 use crate::types::SubgraphDeploymentID;
 
 pub mod monitor;
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, PartialEq, Eq)]
 pub struct Allocation {
     pub id: Address,
     pub status: AllocationStatus,
@@ -31,7 +26,7 @@ pub struct Allocation {
     pub query_fees_collected: Option<U256>,
 }
 
-#[derive(Debug, Eq, PartialEq)]
+#[derive(Clone, Debug, Eq, PartialEq)]
 pub enum AllocationStatus {
     Null,
     Active,
@@ -40,7 +35,7 @@ pub enum AllocationStatus {
     Claimed,
 }
 
-#[derive(Debug, Eq, PartialEq, Deserialize)]
+#[derive(Clone, Debug, Eq, PartialEq, Deserialize)]
 pub struct SubgraphDeployment {
     pub id: SubgraphDeploymentID,
     #[serde(rename = "deniedAt")]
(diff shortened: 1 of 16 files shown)


common/src/allocations/mod.rs
```diff
@@ -94,171 +89,3 @@ impl<'d> Deserialize<'d> for Allocation {
...
-    use std::str::FromStr;
-
-    use crate::prelude::SubgraphDeploymentID;
-
-    use super::*;
-
-    const INDEXER_OPERATOR_MNEMONIC: &str = "abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon abandon about";
-
-    #[test]
-    fn test_derive_key_pair() {
```
hopeyen, line 158: should we keep these unit tests around?

hopeyen (commented):
sweet sweet abstractions with eventuals and network subgraph generic query🤌

common/src/attestations/signers.rs
```diff
@@ -3,162 +3,53 @@
...
-    }
-
-    async fn update_loop(inner: Arc<AttestationSignersInner>) {
-        let mut watch_receiver = inner.allocation_monitor.subscribe();
+) -> Eventual<HashMap<Address, AttestationSigner>> {
+    // Keep a cache of the most recent 1000 signers around so we don't need to recreate them
+    // every time there is a small change in the allocations
+    let cache: &'static Mutex<LruCache<_, _>> = Box::leak(Box::new(Mutex::new(LruCache::new(
+        NonZeroUsize::new(1000).unwrap(),
+    ))));
```
aasseman, line 26: What is the rationale for using an LRU cache rather than updating a hashmap like before? It looks like a roundabout way of achieving a similar thing, but with the added hardcoded limit of 1000 allocations.

aasseman, line 26: What is the rationale for using an LRU cache? Wasn't it overall simpler and more efficient to just update a HashMap in place directly? Especially when you consider that the LRU cache is a souped up HashMap itself.

aasseman (changes requested):
I had a bit of a hard time reviewing because there is a lot of changes that are not strictly related to the implementation of eventuals (refactor of the internals of the signers code, network subgraph), and the removal of many tests, some of which I think could have worked without significant modifications, prevents me from checking that the behavior is correct.

In general I prefer single concern PRs as they are more palatable to the reviewer, and it's easier to rollback PRs if there's a problem in the future without rolling back unrelated changes.

service/src/main.rs
```diff
@@ -61,26 +62,24 @@ async fn main() -> Result<(), std::io::Error> {
     // Make an instance of network subgraph at either
     // graph_node_query_endpoint/subgraphs/id/network_subgraph_deployment
     // or network_subgraph_endpoint
-    let network_subgraph = NetworkSubgraph::new(
+    let network_subgraph: &'static NetworkSubgraph = Box::leak(Box::new(NetworkSubgraph::new(
```
aasseman, line 65: This works, but since NetworkSubgraph is ARC, it could also have been cloned around. Since it indeed is supposed to stick around the whole lifetime of the program it does not technically make a big difference though.

aasseman: Perhaps I should have used this pattern since the beginning though, makes a lot of things even simpler. I suppose the only downside is that it makes everything expect a static reference, which can be sometimes limiting in a library?

Jannis:
@aasseman I can try to pick the changes apart a bit and submit separate PRs. I like more incremental PRs as well. And yes, I need to invest more into testing again.

PR #59, opened 2023-09-29
fix: resolve clippy errors introduced recently

Jannis:
The title says it all. Noticed that these are failing the pipeline on `main`.


Theodus (approved):
LGTM

PR #65, opened 2023-10-11
Refactor to use eventuals for allocations and signers (v2)

Jannis:
This is the final PR to replace #54.

It turns allocation monitoring and attestation signer creation into eventuals, which simplifies the logic and makes it easy to chain different values that change over time together.


aasseman (changes requested):

common/src/attestations/signers.rs
```diff
@@ -3,161 +3,92 @@
...
-    }
 
-    pub async fn read(
-        &self,
-    ) -> tokio::sync::RwLockReadGuard<'_, HashMap<Address, AttestationSigner>> {
-        self.inner.attestation_signers.read().await
-    }
+            HashMap::from_iter(cache.iter().map(|(k, v)| (*k, v.clone())))
+        }
+    })
```
aasseman, line 54: This is just a weak suggestion. No big deal.
```suggestion
) -> Eventual<HashMap<Address, AttestationSigner>> {
    let attestation_signers_map: &'static Mutex<HashMap<Address, AttestationSigner>> =
        Box::leak(Box::new(Mutex::new(HashMap::new())));

    let indexer_mnemonic = Arc::new(indexer_mnemonic);

    // Whenever the indexer's active or recently closed allocations change, make sure
    // we have attestation signers for all of them
    indexer_allocations.map(move |allocations| {
        let indexer_mnemonic = indexer_mnemonic.clone();

        async move {
            let mut attestation_signers_map = attestation_signers_map.lock().await;

            // Remove signers for allocations that are no longer active or recently closed
            attestation_signers_map.retain(|id, _| allocations.contains_key(id));

            // Create signers for new allocations
            for (id, allocation) in allocations.iter() {
                if !attestation_signers_map.contains_key(id) {
                    let signer = AttestationSigner::new(
                        &indexer_mnemonic,
                        allocation,
                        chain_id,
                        dispute_manager,
                    );

                    if let Err(e) = signer {
                        warn!(
                            "Failed to establish signer for allocation {}, deployment {}, createdAtEpoch {}: {}",
                            allocation.id, allocation.subgraph_deployment.id,
                            allocation.created_at_epoch, e
                        );
                    } else {
                        attestation_signers_map.insert(*id, signer.unwrap());
                    }
                }
            }

            attestation_signers_map.clone()
        }
    })
```

common/src/attestations/signer.rs
```diff
@@ -5,48 +5,118 @@ use alloy_primitives::Address;
...
+            if allocation_wallet.address().as_fixed_bytes() == allocation.id {
+                return Ok(allocation_wallet.signer().clone());
+            }
+        }
+    }
+    Err(anyhow::anyhow!(
+        "Could not find allocation signer for allocation {}",
+        allocation.id
+    ))
+}
```
aasseman, line 79: This can be refactored to use `wallet_for_allocation()`

aasseman (changes requested):

common/src/attestations/signer.rs
```diff
@@ -102,21 +172,155 @@ pub struct Attestation {
...
-        deployment_id,
-    );
-    Ok(signer)
+fn wallet_for_allocation(
+    indexer_mnemonic: &str,
+    allocation: &Allocation,
+) -> Result<Wallet<SigningKey>, anyhow::Error> {
+    // Guess the allocation index by enumerating all indexes in the
+    // range [0, 100] and checking for a match
+    for i in 0..100 {
```
aasseman, line 181: Is the allocation index search just a remnant from when we could open parallel allocations? Is it still needed?

Jannis: Yes, exactly! I think opening parallel allocations is still possible and probably still done here and there?

aasseman: I know that it's not supported in the indexer-agent anymore (and that's been for quite a while now IIRC), so supporting it in the indexer-service would only be useful for indexers that use some unsupported method of opening their allocations.
Do you think it'd make sense to propose deprecating it in the indexer-service?

common/src/attestations/signer.rs
Jannis: It's actually unused now. I've updated the tests for it to be tests for `AttestationSigner::new`.

common/src/attestations/signer.rs
Jannis: Let's do that separately. I agree it may be worth doing but it doesn't hurt to keep it in for the moment.

common/src/attestations/signers.rs
Jannis: @aasseman Great! I added this with a small variable name change. You're right, the LRU cache isn't needed since we only ever care about retaining the attestation signers for allocations that exist across two runs of this update logic.

common/src/attestations/signer.rs
aasseman: :100:

aasseman (approved):
:shipit:

PR #66, opened 2023-10-11
Reduce cost model test code

Jannis:
This implements @aasseman's suggestion from https://github.com/graphops/indexer-service-rs/pull/63#discussion_r1355483452.


aasseman (approved):
:shipit:

PR #67, opened 2023-10-12
Type network subgraph deployment strongly

Jannis:
This is based on #65 and will need to be rebased after that has been merged. Only the most recent commit in this PR is actually new.


aasseman (dismissed):
:shipit: (after rebase onto latest main)

aasseman (approved):
:shipit:

PR #68, opened 2023-10-14
Use attestation from toolshed

Theodus:


service/src/query_processor.rs
```diff
@@ -209,30 +203,27 @@ mod tests {
...
-            v: 27,
-            r: hex!("a0c83c0785e2223ac1ea1eb9e4ffd4ca867275469a7b73dab24f39ddcdec5466").into(),
-            s: hex!("4d0457efea889f2ec7ffcc7ff9b408428d0691356f34b01f419f7674d0eb4ddf").into(),
-        };
-
-        assert_eq!(attestation, expected_signature);
+        assert_eq!(
+            attestation_signer.verify(&attestation, request, response, &indexer),
+            Ok(())
+        );
```
aasseman, line 227: This test fails somehow. Also, I think that'd be a little bit less convoluted:
```suggestion
        attestation_signer.verify(&attestation, request, response, &indexer).unwrap();
```

aasseman (changes requested):
LGTM but test is failing

aasseman (approved):
:shipit:

PR #70, opened 2023-10-17
fix: remove irrelevant fields from deployment queries

Jannis:
We may need to add these back later for other purposes, but indexer-service does not need these fields; the only reason it queries allocations is to create attestation signers for them.

There might be even more fields that we could remove. However, signalled_tokens in particular is necessary at the moment because there is buggy data in the real-world network subgraph where the field is negative sometimes. That breaks U256 parsing. Since we don't need this and the other fields, I thought it best to remove them.


hopeyen (approved):
lgtm👍

aasseman (approved):

PR #71, opened 2023-10-17
feat: add dispute manager eventual

Jannis:
This is clearly better than having to define it on the command line. 😁 

Also, the dispute manager _miiiiight_ change over time, so an eventual is nice to make sure we stay on top of that.


aasseman (approved):
:shipit:

PR #72, opened 2023-10-17
Generalize NetworkSubgraph into SubgraphClient

Jannis:
This way we can also it for the escrow subgraph and others that we may need in the future.

This is based on #71 and will need to be rebased when that has been merged.


aasseman (approved):
:shipit:

aasseman (approved):

PR #73, opened 2023-10-17
Turn escrow account monitoring into an eventual in indexer-common

Jannis:
The title says it all 😉


common/src/escrow_accounts.rs
```diff
@@ -0,0 +1,165 @@
...
+            if let Some(errors) = response.errors {
+                warn!(
+                    "Errors encountered fetching escrow accounts for indexer {:?}: {}",
+                    indexer_address,
+                    errors
+                        .into_iter()
+                        .map(|e| e.message)
+                        .collect::<Vec<_>>()
+                        .join(", ")
+                );
```
aasseman, line 77: I'd log this as an error at least, since the indexers would refuse service if the query went wrong.

common/src/escrow_accounts.rs
```diff
@@ -0,0 +1,165 @@
...
+
+                    Ok((account.sender.id, balance))
+                })
+                .collect::<Result<HashMap<_, _>, anyhow::Error>>()
+                .map_err(|e| format!("{}", e))?;
+
+            Ok(sender_accounts)
+        },
+        move |err: String| {
+            warn!(
```
aasseman, line 106: Same here

service/src/tap_manager.rs
```diff
@@ -63,9 +64,11 @@ impl TapManager {
...
-            .escrow_monitor
-            .is_sender_eligible(&receipt_signer)
+            .escrow_accounts
+            .value()
             .await
+            .map(|accounts| accounts.contains_key(&receipt_signer))
+            .unwrap_or(false)
         {
             return Err(QueryError::Other(anyhow::Error::msg(format!(
                 "Receipt's sender ({}) is not eligible for this indexer",
```
aasseman, line 74: This is too permissive. Either the balance should be checked like in https://github.com/graphops/indexer-service-rs/blob/a0b1c876ba7a60b27e8bc63925eed7deae661e1b/service/src/escrow_monitor.rs#L174-L179 or accounts with 0 balance should be filtered out early there https://github.com/graphops/indexer-service-rs/blob/72e44a258bda429f19d499a7854c1a1b2ab6342c/common/src/escrow_accounts.rs#L80-L103

aasseman (changes requested):

common/src/escrow_accounts.rs
Jannis: The way I think about errors is that you should only log errors if they are severe enough for you to get up at night. Everything else is a warning (or something else that's less severe).

That actually reminds me: whenever we encounter issues like this, we should update a prometheus metric (like indexer-service does with `indexerError`, which both logs the error and updates a metric). Will put that on my todo list.

Jannis: What do you think about my reasoning? I'm ok logging an error if you have a different perspective.

Jannis: Actually, I broke my own rule several times in indexer-service. So let's do an `error!` here like you said.

service/src/tap_manager.rs
Jannis: Oh! Good catch. Will fix.

common/src/escrow_accounts.rs
Jannis: Fixed!

common/src/escrow_accounts.rs
Jannis: Fixed!

service/src/tap_manager.rs
Jannis: Fixed as well :)

common/src/escrow_accounts.rs
aasseman: > The way I think about errors is that you should only log errors if they are severe enough for you to get up at night. Everything else is a warning (or something else that's less severe).

The indexer-service would immediately start rejecting all queries. IMO that's quite bad.

> That actually reminds me: whenever we encounter issues like this, we should update a prometheus metric (like indexer-service does with indexerError, which both logs the error and updates a metric). Will put that on my todo list.

:100:

aasseman: > > The way I think about errors is that you should only log errors if they are severe enough for you to get up at night. Everything else is a warning (or something else that's less severe).
> 
> The indexer-service would immediately start rejecting all queries. IMO that's quite bad.

Actually, after putting in some more thought, it's slightly more subtle than that, I think. Here it will just log the errors and move on, so at worst it will yield an empty list of accounts:
https://github.com/graphops/indexer-service-rs/blob/b567f9a28d35a674a24d4d0d30cabcfa1c1663ea/common/src/escrow_accounts.rs#L67-L78
That would result in the indexer-service refusing all queries.
Whereas here it will retry a few times, and perhaps just keep the old list of accounts?
https://github.com/graphops/indexer-service-rs/blob/b567f9a28d35a674a24d4d0d30cabcfa1c1663ea/common/src/escrow_accounts.rs#L67-L78
The risk here is that the indexer-service serves queries to escrow accounts that got emptied. But it should get caught by the TAP agent before going over the maximal admissible loss.

Overall, I'm not quite sure what the best solution is here :shrug:

aasseman (approved):

PR #78, opened 2023-10-20
fix(service): ms intervals treated as seconds

aasseman:
Because the indexer-service arguments are asking for ms:
https://github.com/graphops/indexer-service-rs/blob/b603c306969e1a202ea15381438cde3a5204dcc3/service/src/config.rs#L211-L218
https://github.com/graphops/indexer-service-rs/blob/b603c306969e1a202ea15381438cde3a5204dcc3/service/src/config.rs#L260-L267
(I'll fix clap for this second one later)

This is carried from the original `indexer-service`. Even though asking for seconds directly may make more sense, that would introduce a breaking change for users.

### DIFF
--- a/service/src/main.rs
+++ b/service/src/main.rs
@@ -81,7 +81,7 @@ async fn main() -> Result<(), std::io::Error> {
         network_subgraph,
         config.ethereum.indexer_address,
         1,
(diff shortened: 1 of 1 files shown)


hopeyen (approved):
🔍

PR #81, opened 2023-10-23
Properly implement subgraph client fallback from local deployment to remote URL

Jannis:
~This may deviate from the existing indexer-service, but that one has grown over time. Really, even if you index the network subgraph deployment (or escrow subgraph) locally, you just need a query URL in the indexer service. The optional deployment logic is handled by indexer-agent.~

This now implements the correct logic to monitor the local subgraph deployment status and, if it's not healthy and synced, falls back to the remote URL provided. This mimics the behavior of the existing indexer service: https://github.com/graphprotocol/indexer/blob/main/packages/indexer-common/src/network-subgraph.ts#L143-L167.


aasseman:
Makes sense! Here's how I understood it:
- The indexer-agent is tasked with deploying the network subgraph to the graph-node + managing fallback to a public URL.
- The indexer-service would only need to point to the indexer-agent.

If I understood it correctly, then shouldn't that be extended further, and the SubgraphClient only point to a user-provided `INDEXER_AGENT_QUERY_ENDPOINT` (or similar)?

aasseman:
PS: and the actual path to the correct subgraph endpoint in indexer-agent would be hardcoded, ie. something like:
```
let indexer_agent_subgraphs_endpoint = "http://localhost:8080/subgraphs/"
let network_subgraph = SubgraphClient::from_something(indexer_agent_subgraphs_endpoint, "escrow-subgraph");
```

common/src/subgraph_client.rs
```diff
@@ -20,48 +19,32 @@ pub struct SubgraphClient {
...
-        // TODO: Check indexing status of the local subgraph deployment
-        // if the deployment is healthy and synced, use local_subgraoh_endpoint
-        let _local_subgraph_endpoint = match (graph_node_query_endpoint, deployment) {
-            (Some(endpoint), Some(id)) => Some(Self::local_deployment_endpoint(endpoint, id)?),
-            _ => None,
-        };
-
-        let subgraph_url = Url::parse(subgraph_url)
-            .map_err(|e| anyhow!("Could not parse subgraph url `{}`: {}", subgraph_url, e))?;
+    pub fn new(name: &str, query_url: &str) -> Result<Self, anyhow::Error> {
```
aasseman, line 22: Do we really need `name`? It seems to be only used for error messages.

Jannis:
> Makes sense! Here's how I understood it:
> 
> * The indexer-agent is tasked with deploying the network subgraph to the graph-node + managing fallback to a public URL.
> * The indexer-service would only need to point to the indexer-agent.
> 
> If I understood it correctly, then shouldn't that be extended further, and the SubgraphClient only point to a user-provided `INDEXER_AGENT_QUERY_ENDPOINT` (or similar)?

I re-checked what the existing indexer-service does and it's a bit more complicated. It essentially polls the status of the local deployment and uses that if synced, otherwise it uses the remote subgraph. I think we need at least a local and remote URL for the network subgraph (same with escrow ideally). I'll rework this PR to match the existing behavior.

common/src/subgraph_client.rs
Jannis: Yeah, no, let's not include it.

aasseman:
> > Makes sense! Here's how I understood it:
> > 
> > * The indexer-agent is tasked with deploying the network subgraph to the graph-node + managing fallback to a public URL.
> > * The indexer-service would only need to point to the indexer-agent.
> > 
> > If I understood it correctly, then shouldn't that be extended further, and the SubgraphClient only point to a user-provided `INDEXER_AGENT_QUERY_ENDPOINT` (or similar)?
> 
> I re-checked what the existing indexer-service does and it's a bit more complicated. It essentially polls the status of the local deployment and uses that if synced, otherwise it uses the remote subgraph. I think we need at least a local and remote URL for the network subgraph (same with escrow ideally). I'll rework this PR to match the existing behavior.

Oh, ok then!

common/src/subgraph_client/client.rs
```diff
@@ -0,0 +1,202 @@
...
+            .await
+            .and_then(|response| response.error_for_status())?
+            .json::<Response<T>>()
+            .await?)
+    }
+}
+
+/// Network subgraph query wrapper
+///
+/// This is Arc internally, so it can be cloned and shared between threads.
```
aasseman, line 89: I think it's not strictly the case anymore.

aasseman (approved):
LGTM. Just one comment in the code that may be obsolete.

common/src/subgraph_client/client.rs
Jannis: Addressed!

Jannis:
Added a bunch more tests for the fallback behavior. This should be good to go now.

aasseman (approved):
:shipit:

PR #82, opened 2023-10-24
fix: add config for receipts domain

Theodus:
This adds the config options, `chain_id` & `verifying_contract`, required to contruct the EIP-712 domain used when verifying receipts.


hopeyen (approved):
lgtm! Could you please leave some documentation on available contract addresses one could use, perhaps add a default?

PR #86, opened 2023-11-07
fix: fail fast when escrow accounts are missing

Theodus:
Previsously, paid queries would stall until timeout while awaiting the initial eventual value for escrow accounts. This change avoids the delay, such that the intended error message is returned.

### DIFF
--- a/common/src/tap_manager.rs
+++ b/common/src/tap_manager.rs
@@ -68,14 +68,10 @@ impl TapManager {
             })?;
         if !self
             .escrow_accounts
-            .value()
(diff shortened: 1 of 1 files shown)


aasseman (approved):
:shipit: 
:thinking: I suppose I should have used `value_immediate` in most places where I used `value`...

Theodus:
It seems with the org change I don't have authority to merge

aasseman:
> It seems with the org change I don't have authority to merge

Me neither :grimacing:

PR #87, opened 2023-11-08
fix: share HTTP client between `DeploymentClient`s

Theodus:
Ideally, a single `reqwest::Client` is reused so that they use the same connection pool. This also provides a clearer point to set timeout policies, etc. and removes an unused error case.


hopeyen (approved):
good point!

Issue #89, opened 2023-11-08
incorrect subgraph indexing status check


Theodus:
The behavior of `monitor_deployment_status` is incorrect for the following reasons:
1. The query uses the wrong filter key, resulting in the expected 0 index value potentially being the status for a different deployment. It should be the following:
    ```graphql
    query indexingStatuses($ids: [ID!]!) {
        indexingStatuses(subgraphs: $ids) {
            synced
            health
        }
    }
    ```
2. The `status_url` is not the correct URL. On my local-network, it's set to `http://172.17.0.1:8000/status` instead of `http://localhost:8030/graphql`, despite the latter being set as `--graph-node-status-endpoint`.
    ![image](https://github.com/graphprotocol/indexer-rs/assets/9407624/078cd91e-ad2c-4012-8366-9df0b94a7481)

hopeyen:
Hi Theo, thanks for the detail report!

Trying to understand the issue with

> 2. The `status_url` is not the correct URL. On my local-network, it's set to `http://172.17.0.1:8000/status` instead of `http://localhost:8030/graphql`, despite the latter being set as `--graph-node-status-endpoint`.

indexer service exposes a `/status` endpoint for indexing statuses queries, it basically just routes the queries to `--graph-node-status-endpoint`. If you curl to `http://172.17.0.1:8000/status`, it should give you the same responses. Could you help me understand the issue a bit better?

Theodus:
It's possible that I'm misunderstanding how this should work, but one gives the expected response, the other just returns `{"message": "Not found"}` (using graph-node v0.33.0):
![image](https://github.com/graphprotocol/indexer-rs/assets/9407624/6787dc37-8791-4bd2-bdfb-d3cfc7040943)

hopeyen:
I think I understood the problem, pls lmk if https://github.com/graphprotocol/indexer-rs/pull/91 fixes the problem with `monitor_deployment_status`; this doesn't change anything for the curl request to `localhost/status`, even though that should've worked for indexer service endpoint🤔

PR #90, opened 2023-11-08
fix: indexing statuses filter key

hopeyen:
Fix filter key for `monitor_deployment_status` 
For https://github.com/graphprotocol/indexer-rs/issues/89


Theodus (approved):
LGTM

PR #105, opened 2023-12-01
feat: update tap, toolshed and ethers dependencies

Jannis:
This bumps a number of important dependencies.


aasseman:
Adding the `arbitrary-precision` feature to `serde` will fix the test.
Don't know why it wasn't a problem before :shrug: .

Jannis:
> Adding the `arbitrary-precision` feature to `serde` will fix the test. Don't know why it wasn't a problem before 🤷 .

@aasseman That did the trick indeed!

aasseman (approved):

PR #109, opened 2024-01-09
feat(indexer-service): Sender denylist

aasseman:
indexer-service will reject all paid queries from denied senders. Denylist DB table updates are watched through PG notifications.

### DIFF
--- a/common/src/tap_manager.rs
+++ b/common/src/tap_manager.rs
@@ -1,17 +1,19 @@
 // Copyright 2023-, GraphOps and Semiotic Labs.
 // SPDX-License-Identifier: Apache-2.0
 
+use crate::prelude::Allocation;
 use alloy_primitives::Address;
 use alloy_sol_types::Eip712Domain;
 use anyhow::anyhow;
 use ethers_core::types::U256;
 use eventuals::Eventual;
+use sqlx::postgres::PgListener;
 use sqlx::{types::BigDecimal, PgPool};
+use std::collections::HashSet;
 use std::{collections::HashMap, str::FromStr, sync::Arc};
 use tap_core::tap_manager::SignedReceipt;
+use tokio::sync::RwLock;
 use tracing::error;
-
-use crate::prelude::Allocation;
 
 #[derive(Clone)]
 pub struct TapManager {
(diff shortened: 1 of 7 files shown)


common/src/tap_manager.rs
```diff
@@ -106,6 +143,84 @@ impl TapManager {
...
+            tg_op: String,
+            sender_address: Address,
+        }
+
+        loop {
+            let pg_notification = pglistener.recv().await.expect(
+                "should be able to receive Postgres Notify events on the channel \
+                'scalar_tap_deny_notification'",
+            );
+
```
gusinacio, line 184: Ideally, you would have a select! {} to handle cancelations. Your task should have a CancellationToken that you can `break` the loop so you can safely do a `handle.join()` on Drop.

gusinacio: This might not be required for this task since we don't need to finish the computation before shutting the task down. But it can be considered a good practice to handle all JoinHandle.

aasseman: Yeah I don't think it's super critical here. Nevertheless I took a stab at making it a bit cleaner in 8863e0a1d7e5b3b8b0b29527f05c0b452bdbac7d

aasseman:
Making it draft until #110 gets merged

common/src/tap_manager.rs
```diff
@@ -1,38 +1,72 @@
...
-
 #[derive(Clone)]
 pub struct TapManager {
     indexer_allocations: Eventual<HashMap<Address, Allocation>>,
     escrow_accounts: Eventual<HashMap<Address, U256>>,
     pgpool: PgPool,
     domain_separator: Arc<Eip712Domain>,
+    sender_denylist: Arc<RwLock<HashSet<Address>>>,
+    _sender_denylist_watcher_handle: Arc<tokio::task::JoinHandle<()>>,
+    sender_denylist_watcher_cancel_token: Arc<tokio_util::sync::CancellationToken>,
```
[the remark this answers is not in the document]
aasseman: :facepalm:

gusinacio (approved):
LGTM

gusinacio (approved):
LGTM

PR #110, opened 2024-01-13
fix: support sender signers

aasseman:
The TAP implementation was assuming that the receipts/RAVs are signed by the Sender (aka. Escrow account owner), but they are actually signed by one of the signers (a key specifically created only for signing receipts/RAVs in the sender's name).
A sender can have any number of valid/thawing/expired signers.

This makes the code even uglier... I hope that I can think of a cleaner architecture later.


common/src/escrow_accounts.rs
```diff
@@ -38,50 +68,109 @@ pub fn escrow_accounts(
...
+            }
+        }
+    "#;
+
+    let query = if reject_thawing_signers {
+        query_no_thawing_signers
+    } else {
+        query_with_thawing_signers
+    };
+
```
gusinacio, line 123: You could you the expression directly:
```
let query = if reject_thawing_signers {
    r#"..."#
} else {
    r#"..."#
};

File: tap-agent/src/tap/receipt_storage_adapter.rs [hunk 2]
```diff
@@ -100,17 +103,24 @@ impl ReceiptStorageAdapterTrait for ReceiptStorageAdapter {
...
+            .map_err(|e| AdapterError::AdapterError {
+                error: format!("{:?}.", e),
+            })?;
+
         let records = sqlx::query!(
             r#"
                 SELECT id, receipt
                 FROM scalar_tap_receipts
-                WHERE allocation_id = $1 AND sender_address = $2 AND $3::numrange @> timestamp_ns
+                WHERE allocation_id = $1 AND signer_address IN (SELECT unnest($2::text[]))
```
gusinacio, line 116: Do you need SELECT here?

gusinacio (changes requested):
Great work, I have a few questions about how the contract works, a small suggestion with naming, and another one to remove an unused variable declaration.

common/src/escrow_accounts.rs
```diff
@@ -13,18 +13,48 @@ use tracing::{error, warn};
 
 use crate::prelude::{Query, SubgraphClient};
 
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct EscrowAccounts {
+    pub balances: HashMap<Address, U256>,
+    pub signers_to_senders: HashMap<Address, Address>,
```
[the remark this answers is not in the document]
aasseman: No. It's very important that there is only a single sender behind a signer, so that we always know without ambiguity who to charge just from the receipt signature.
It is enforced by the contracts there:
- https://github.com/semiotic-ai/timeline-aggregation-protocol-contracts/blob/cbc9a5acd3401700bf482ccc24ad05ab339614cd/src/Escrow.sol#L45-L47
- https://github.com/semiotic-ai/timeline-aggregation-protocol-contracts/blob/cbc9a5acd3401700bf482ccc24ad05ab339614cd/src/Escrow.sol#L505

common/src/escrow_accounts.rs
aasseman: Indeed

common/src/escrow_accounts.rs
```diff
@@ -13,18 +13,48 @@ use tracing::{error, warn};
 
 use crate::prelude::{Query, SubgraphClient};
 
+#[derive(Clone, Debug, Default, PartialEq, Eq)]
+pub struct EscrowAccounts {
+    pub balances: HashMap<Address, U256>,
```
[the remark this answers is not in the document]
aasseman: 229b7929f00928f110e8262f127529f3edb5862b

tap-agent/src/tap/escrow_adapter.rs
```diff
@@ -45,14 +45,29 @@ impl EscrowAdapterTrait for EscrowAdapter {
...
+                        "Sender {} not found for receipt signer, could not get available escrow.",
+                        sender
+                    )
+                    .to_string(),
+                })?;
+
+        let balance = escrow_accounts
+            .senders_balances
+            .get(sender)
             .ok_or(AdapterError::AdapterError {
```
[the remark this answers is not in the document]
gusinacio: We could also use `Result` and just use `?`. It makes it more readable where it's being used.

common/src/tap_manager.rs
```diff
@@ -67,11 +67,21 @@ impl TapManager {
...
+            .ok_or_else(|| {
+                anyhow!(
+                    "Receipt signer `{}` is not eligible for this indexer",
+                    receipt_signer
+                )
+            })?;
+
+        if !escrow_accounts
+            .senders_balances
+            .get(receipt_sender)
```
gusinacio, line 84: Another use of `get_balance_for_sender()`

tap-agent/src/tap/escrow_adapter.rs
```diff
@@ -74,16 +89,37 @@ impl EscrowAdapterTrait for EscrowAdapter {
...
+            escrow_accounts
+                .signers_to_senders
+                .get(&sender)
+                .ok_or(AdapterError::AdapterError {
+                    error: format!(
+                        "Sender {} not found for receipt signer, could not get available escrow.",
+                        sender
+                    )
+                    .to_string(),
+                })?;
```
gusinacio, line 119: I think I'd like to have a `get_sender_for_signer(signer: Address) -> Result<Address>` because it hides the data structure used and also we don't need to do `ok_or()` all over the place.

common/src/escrow_accounts.rs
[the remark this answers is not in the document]
gusinacio: That makes sense, thanks!

tap-agent/src/tap/receipt_checks_adapter.rs
```diff
@@ -125,8 +126,11 @@ impl ReceiptChecksAdapterTrait for ReceiptChecksAdapter {
                 })?;
 
         Ok(escrow_accounts
+            .signers_to_senders
             .get(&sender_id)
-            .map_or(false, |balance| *balance > U256::from(0)))
+            .and_then(|sender| escrow_accounts.senders_balances.get(sender))
+            .map(|balance| *balance > U256::from(0))
```
gusinacio, line 132: Another use of `get_balance_for_signer()`

gusinacio (changes requested):

common/src/escrow_accounts.rs
tmigone: god bless rust expressions

common/src/escrow_accounts.rs
```diff
@@ -13,18 +16,84 @@ use tracing::{error, warn};
...
+                sender
+            ))
+            .copied()
+    }
+
+    pub fn get_balance_for_signer(&self, signer: &Address) -> Result<U256> {
+        self.get_sender_for_signer(signer)
+            .and_then(|sender| self.get_balance_for_sender(&sender))
+            .map_err(|e| anyhow::format_err!("Could not get balance for signer {}: {}", signer, e))
+    }
```
gusinacio, line 78: Can we return an `AdapterError` for all of these?

Is there any reason to map to anyhow?

aasseman: Yes because it's not part of the adapters code, and is used outside of adapters in some places. Though it's really only a choice, nothing stops us for throwing `AdapterError`s anyway, what do you think?

gusinacio: I'd rather use AdapterError here. All the code would be much cleaner, also, I'm not sure why `thiserror` is not being so used. 

This could be a different error like `EscrowAccountsError` and used by thiserror to convert to AdapterError automatically.

This would address your concern about not being an AdapterError while helping with ergonomics to remove the 'map_err()'

gusinacio (approved):
Beautiful code. LGTM

PR #112, opened 2024-01-18
Refactor rav requester

aasseman:
It's smaller than it looks, most of the diff is about splitting `rav_quester_try()` into multiple functions to help with readability.


gusinacio (approved):
I like that the rav_requester_task is now a long-running task and it's not being spawned on-demand.

LGTM

PR #114, opened 2024-02-01
fix(common): subgraph monitor query

aasseman:
The subgraph client monitor was sending malformed queries to graph-node.

subgraph monitor logs:
```
WARN indexer_common::subgraph_client::monitor: Error querying deployment status for `QmP9foCa7YRvQJCtrdPV1zu1yVovwagEpcDSarvmQKS642`: Failed to query status of deployment `QmP9foCa7YRvQJCtrdPV1zu1yVovwagEpcDSarvmQKS642`: Error deserializing GraphQL response body: expected value at line 1 column 1
```

Graph-node logs:
```
ERRO IndexNodeService call failed: GraphQL server error (query error): query parse error: Parse error at 1:2\nUnexpected `"query"[StringValue]`\nExpected `Name`\n, component: IndexNodeServer
```

Changing the query body to use `indexer_common::subgraph_client::client:Query` fixes the problem.

I also fixed a small inconsequential typo: graph-node's status GQL API actually expects the deployment IDs to be of type `String`, and not `ID`, according to the introspected schema.

Tested against the `local-network`.

### DIFF
--- a/common/src/subgraph_client/monitor.rs
+++ b/common/src/subgraph_client/monitor.rs
@@ -4,13 +4,18 @@
 use std::time::Duration;
 
 use eventuals::{timer, Eventual, EventualExt};
-use graphql_http::http_client::{ReqwestExt, ResponseResult};
+use graphql_http::{
+    http::request::IntoRequestParameters,
+    http_client::{ReqwestExt, ResponseResult},
+};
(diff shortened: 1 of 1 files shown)


gusinacio (approved):
LGTM

PR #118, opened 2024-02-05
build(deps): fix toolshed deps

aasseman:
Fixed the versions of `toolshed` crates to proper tags.
Also, replaced direct usage of `alloy_primitives::Address` to the toolshed re-export `thegraph::types::Address` to enforce `alloy` version compatibility between `indexer-rs` and other projects that depend on `toolshed`.

Note that the `tap_core` dependency is based on a custom branch based off of `tap_core` v0.7.0 on which the same treatment was applied. This is temporary as for now `tap_core`'s main branch contains a lot of breaking changes.


hopeyen (approved):
lg, thanks! :ship:

PR #120, opened 2024-02-07
feat(common): add a route /info for operator public key

hopeyen:
adding `/info` endpoint back to query operator info

### DIFF
--- a/common/src/address.rs
+++ b/common/src/address.rs
@@ -1,0 +1,21 @@
+// Copyright 2023-, GraphOps and Semiotic Labs.
+// SPDX-License-Identifier: Apache-2.0
+
+use ethers::signers::{
+    coins_bip39::English, LocalWallet, MnemonicBuilder, Signer, Wallet, WalletError,
+};
+use ethers_core::k256::ecdsa::SigningKey;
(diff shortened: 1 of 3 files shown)


aasseman (changes requested):
Perhaps could the operator address be stored as a lazy static instead of recomputing it every time?
Otherwise LGTM

hopeyen:
> Perhaps could the operator address be stored as a lazy static instead of recomputing it every time?

Operator address computed once at the initialization stage of IndexerService, and simply returned when it gets queried. I'm not sure storing operator address as lazy_static since it should be computed based on operator mnemonic. Could you explain a bit more please?

aasseman:
Ah yes, you're right, I misread it!

aasseman (approved):

Jannis:
@hopeyen This PR is incompatible with the `/info` handler of the existing indexer service, which returns an object. Could we fix that and make them compatible?

https://github.com/graphprotocol/indexer/blob/e701da80600e91ae7c89e2ca0962aea40360082e/packages/indexer-service/src/server/operator.ts#L13

PR #123, opened 2024-02-21
fix(tap-agent): make RAV table sequelize-friendly

aasseman:
By adding the `createdAt` and `updatedAt` columns. 
Sequelize is used in indexer-agent, and it needs to access the RAV table to redeem RAVs.

See https://sequelize.org/docs/v6/core-concepts/model-basics/#timestamps


gusinacio (approved):
LGTM

PR #124, opened 2024-02-26
fix: use 65 bytes signature

gusinacio:
remove open-fastrlp dependency, using bytes directly


gusinacio:
We'll use rlp inside indexer-agent

aasseman (approved):

aasseman (approved):

PR #126, opened 2024-02-26
fix: capitalize scalar receipt header

hopeyen:
I was having trouble receiving scalar receipts, so took a look at how the gateway construct the receipt, found that it creates receipt's header with capital letters `HeaderName::from_str("Scalar-Receipt").unwrap(),` from [line](https://github.com/edgeandnode/graph-gateway/blob/f32e5f33447d27f6425fa2754f64bc1b5acfe1ef/graph-gateway/src/indexer_client.rs#L66)
indexer framework process header with just lower-case `scalar-receipt`. I *think* the convention uses capital letters so proposing a change here. if not, I can propose a change on the gateway's side.

### DIFF
--- a/common/src/indexer_service/http/scalar_receipt_header.rs
+++ b/common/src/indexer_service/http/scalar_receipt_header.rs
@@ -25,7 +25,7 @@ impl Deref for ScalarReceipt {
 }
 
 lazy_static! {
-    static ref SCALAR_RECEIPT: HeaderName = HeaderName::from_static("scalar-receipt");
+    static ref SCALAR_RECEIPT: HeaderName = HeaderName::from_static("Scalar-Receipt");
 }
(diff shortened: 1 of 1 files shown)


aasseman:
Indeed. Though what's weird is that we haven't seen any issues in integration testing with this :thinking:.

hopeyen:
> Indeed. Though what's weird is that we haven't seen any issues in integration testing with this 🤔.

right? very weird, I found this issue while testing with using indexer-rs in file service. updating this is what allowed the file provider to realize there is a receipt (though still having trouble at the next step of verify_and_store_receipt, ignore that for now)

aasseman:
Yup the TAP bits are still being polished

aasseman (approved):

PR #130, opened 2024-03-01
fix: wait to retry on rav request error

gusinacio:
Signed-off-by: Gustavo Inacio <gustavo@semiotic.ai>

### DIFF
--- a/tap-agent/src/tap/sender_account.rs
+++ b/tap-agent/src/tap/sender_account.rs
@@ -48,7 +48,14 @@ impl Inner {
         loop {
             notif_value_trigger.notified().await;
 
-            self.rav_requester_single().await;
+            if let Err(error) = self.rav_requester_single().await {
+                // If an error occoured, we shouldn't retry right away, so we wait for a bit.
+                error!(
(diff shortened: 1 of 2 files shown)


tap-agent/src/tap/sender_account.rs
```diff
@@ -48,7 +48,14 @@ impl Inner {
         loop {
             notif_value_trigger.notified().await;
 
-            self.rav_requester_single().await;
+            if let Err(error) = self.rav_requester_single().await {
+                // If an error occoured, we shouldn't retry right away, so we wait for a bit.
```
aasseman, line 52: I think the code for doing the actual waiting is missing? (sleep?)

aasseman (changes requested):
Also fmt error

tap-agent/src/tap/sender_account.rs
gusinacio: we don't need to sleep (at least for testing) because it will wait for it to be notified again by other receipt.

But indeed we could add a sleep backoff in case of a high rate of receipts.

aasseman (approved):

PR #150, opened 2024-04-17
fix: use closedAt timestamp instead of epoch

gusinacio:


common/src/allocations/monitor.rs
```diff
@@ -111,12 +81,13 @@ pub fn indexer_allocations(
...
-                .map_err(|e| format!("Failed to fetch current epoch: {}", e))?;
-
-            // Allocations can be closed one epoch into the past
-            let closed_at_epoch_threshold = current_epoch - 1;
+            // Allocations can be closed 5 mins into the past
+            let start = SystemTime::now();
+            let since_the_epoch = start
+                .duration_since(UNIX_EPOCH)
+                .expect("Time went backwards");
+            const MINUTES_THRESHOLD: u64 = 5;
```
aasseman, line 89: Perhaps are safer default could be 1 hour? (yes sorry I know I gave the example of 5min earlier today).
Could it make sense to have this a program parameter?

gusinacio: Well, in that case we'd need to have a parameter for both indexer-service and tap-agent.

gusinacio: Just added a config for both indexer-service and tap-agent. The default value is 1 hour.

tap-agent/src/agent.rs
```diff
@@ -76,8 +74,8 @@ pub async fn start_agent() -> (ActorRef<SenderAccountsManagerMessage>, JoinHandl
     let indexer_allocations = indexer_allocations(
         network_subgraph,
         *indexer_address,
-        GRAPH_NETWORK_ID,
         Duration::from_millis(*allocation_syncing_interval_ms),
+        Duration::from_millis(*recently_closed_allocation_buffer_seconds),
```
aasseman, line 78:
```suggestion
        Duration::from_secs(*recently_closed_allocation_buffer_seconds),
```

common/src/allocations/monitor.rs
```diff
@@ -111,12 +82,12 @@ pub fn indexer_allocations(
     // Refresh indexer allocations every now and then
     timer(interval).map_with_retry(
         move |_| async move {
-            let current_epoch = current_epoch(network_subgraph, graph_network_id)
-                .await
-                .map_err(|e| format!("Failed to fetch current epoch: {}", e))?;
-
-            // Allocations can be closed one epoch into the past
-            let closed_at_epoch_threshold = current_epoch - 1;
+            // Allocations can be closed 5 mins into the past
```
aasseman, line 85:
```suggestion
            // Allocations are eligible even if closed for up to `recently_closed_allocation_buffer` 
```

aasseman (changes requested):
LGTM, just a couple little things

aasseman (approved):

PR #151, opened 2024-04-18
Gusinacio/bump axum reqwest

gusinacio:
This PR aims to update `graphql-http` into the released version of `thegraph-graphql-http`.

By doing it, I had to update `reqwest` to version 0.12 which triggered a whole chain effect of updating multiple dependencies including: `axum`, add `axum-extra`, `tower`, `tower-http`, `tower-governor`, `autometrics`, `async-graphql` and `async-graphql-axum`.

The rest of the changes are just fixes for the breaking changes of each dependency.

The reason for the update of `thegraph-graphql-http` is that we are getting some serialization errors and we plan to fix them soon on a future pr.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -388,49 +388,16 @@ dependencies = [
 
 [[package]]
 name = "async-graphql"
-version = "6.0.11"
+version = "7.0.3"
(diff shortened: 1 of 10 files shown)


aasseman (approved):
LGTM

PR #153, opened 2024-04-19
feat: add value check

gusinacio:
This PR aims to create the value check for indexer-service. This means that our system checks if the gateway is sending the value that we are requesting, otherwise it rejects the query.

Initially, I was planning to do a cache system to allow queries that contain the older cost model but it exploded really fast in complexity.

Given conversations with indexers, they seemed to not care to lose a few queries during 30 seconds until the gateway fetches again the cost model.

In case this becomes a problem to indexers, we can implement it again.


common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,271 @@
...
+            expected_value,
+            should_accept,
+            "Evaluating mininum query fee."
+        );
+
+        if should_accept {
+            Ok(())
+        } else {
+            return Err(anyhow!(
+                "Query receipt does not have the minimum value. Expected value: {}. Minimum value: {}.",
```
aasseman, line 185:
```suggestion
                "Query receipt does not have the minimum value. Expected value: {}. Received value: {}.",
```

common/src/indexer_service/http/request_handler.rs
```diff
@@ -41,16 +44,40 @@ where
...
+    }
+
     let request =
         serde_json::from_slice(&body).map_err(|e| IndexerServiceError::InvalidRequest(e.into()))?;
 
     let mut attestation_signer: Option<AttestationSigner> = None;
 
     if let Some(receipt) = receipt.into_signed_receipt() {
         let allocation_id = receipt.message.allocation_id;
+        let signature = receipt.signature;
```
aasseman, line 60: I don't understand how/why you're using the signature

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,271 @@
...
+
+    fn get_expected_value(&self, query_id: &Signature) -> anyhow::Result<u128> {
+        // get query from key
+        let agora_query = self
+            .get_agora_query(query_id)
+            .ok_or(anyhow!("No query found"))?;
+
+        // get agora model for the allocation_id
+        let mut cache = self.cost_model_cache.lock().unwrap();
+        // on average, we'll have zero or one model
```
aasseman, line 150: I'd expect that we can get more than 1 model in the case the model was recently updated, and we are still in the period where expect the query to be priced according to a previous model or the new one.

aasseman:
Having the cache in the indexer-service raises a small risk though.
We treat indexer-service as a stateless service. So in the case an indexer-service was recently started, it wouldn't be aware of previous recent agora models, and that could lead to wrongly rejecting queries.
However the only other solution to this would be to have the indexer DB do the caching of models, which could be annoying to implement.

common/src/tap/checks/value_check.rs
gusinacio: indeed, we'll have more than one model when recently updated. But I expect that this is not that often. That's the reason for my comment. On average we'll have one model, or even zero if it's not set.

common/src/indexer_service/http/request_handler.rs
gusinacio: Signature is the best "unique id" for a specific receipt. It is unique for a given receipt + signer.

aasseman: I thought you wanted it to be unique per "SQL text" vs. unique per each query.

common/src/cost_model.rs
```diff
@@ -0,0 +1,465 @@
...
+
+        // If we have a cost model and a global cost model, merge them
+        (Some(model), Some(global_model)) => Some(merge_global(model, &global_model)),
+
+        // If we have only a global model, use that
+        (None, Some(global_model)) => Some(CostModel {
+            deployment: deployment.to_owned(),
+            model: global_model.model,
+            variables: global_model.variables,
+        }),
```
suchapalaver, line 162: I'm curious what made you choose the order here. Is `None, None` the most likely path?

common/src/cost_model.rs
```diff
@@ -0,0 +1,465 @@
...
+                    .parse()
+                    .unwrap(),
+                model: Some("default => 0.00012;".to_string()),
+                variables: None,
+            },
+        ]
+    }
+
+    #[sqlx::test(migrations = "../migrations")]
+    async fn success_cost_models(pool: PgPool) {
```
suchapalaver, line 254: Naming question: Could this test name give a better sense of what it does?

common/src/cost_model.rs
gusinacio: TBH, I'm just moving this code to common, it was previously on `service/src/database.rs`

common/src/cost_model.rs
```diff
@@ -0,0 +1,465 @@
...
+fn merge_global(model: CostModel, global_model: &DbCostModel) -> CostModel {
+    CostModel {
+        deployment: model.deployment,
+        model: model.model.clone().or(global_model.model.clone()),
+        variables: model.variables.clone().or(global_model.variables.clone()),
+    }
+}
+
+#[cfg(test)]
+pub(crate) mod test {
```
suchapalaver, line 191: I understand this isn't a focus of this PR but this test suite could benefit from more defensive tests rather than just focusing on what we'd expect given the inputs we selected.

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+
+use tap_core::receipt::{
+    checks::{Check, CheckError, CheckResult},
+    state::Checking,
+    Context, ReceiptWithState,
+};
+
+use crate::cost_model;
+
+// we only accept receipts with minimal 1 wei grt
```
suchapalaver, line 27:
```suggestion
/// We only accept receipts with minimal 1 wei grt
```

?

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+            global_model,
+            cost_model_map,
+            watcher_cancel_token,
+        }
+    }
+
+    fn get_expected_value(&self, agora_query: &AgoraQuery) -> anyhow::Result<u128> {
+        // get agora model for the deployment_id
+        let model = self.cost_model_map.read().unwrap();
+        let subgraph_model = model.get(&agora_query.deployment_id);
```
suchapalaver, line 194:
```suggestion
        let model_for_deployment_id = self.cost_model_map.read().unwrap();
        let subgraph_model = model_for_deployment_id.get(&agora_query.deployment_id);
```

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+
+    #[sqlx::test(migrations = "../migrations")]
+    async fn should_watch_model_insert(pgpool: PgPool) {
+        let check = MinimumValue::new(pgpool.clone()).await;
+        assert_eq!(check.cost_model_map.read().unwrap().len(), 0);
+
+        // insert 2 cost models for different deployment_id
+        let test_models = crate::cost_model::test::test_data();
+        add_cost_models(&pgpool, to_db_models(test_models.clone())).await;
+        sleep(Duration::from_millis(200)).await;
```
suchapalaver, line 354: The `sleep` makes things a bit confusing.

You could use something like this:

```rust
async fn wait_for_update(check: &MinimumValue) {
    tokio::time::timeout(Duration::from_secs(1), async {
        while check.cost_model_map.read().unwrap().len() != expected_len {
            sleep(Duration::from_millis(50)).await;
        }
    }).await.unwrap();
}
```

I think the intention is clearer.

suchapalaver (approved):
This looks good to me @gusinacio , a lot of work in the making!

I think the tests could be made more defensive and clearer about intention, but I understand some of the limitations we're facing with testing and I trust this has been tested extensively observationally.

common/src/cost_model.rs
```diff
@@ -0,0 +1,465 @@
...
+            deployment: format!("{deployment:#x}"),
+            model: model.model,
+            variables: model.variables,
+        }
+    }
+}
+
+/// Query cost models from the database, merging the global cost model in
+/// whenever there is no cost model defined for a deployment.
+pub async fn cost_models(
```
anirudh2, line 55: In my opinion, this function does way too many different things. I recommend creating helper functions such as:

* `convert_deployments_to_hex`
* `query_cost_models`
* `filter_deployments`
* `merge_cost_model`

common/src/cost_model.rs
```diff
@@ -0,0 +1,465 @@
...
+                    }),
+            )
+            .collect();
+    }
+
+    Ok(models)
+}
+
+/// Make database query for a cost model indexed by deployment id
+pub async fn cost_model(
```
anirudh2, line 128: Same thing. I think this function does too many different things.

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+    state::Checking,
+    Context, ReceiptWithState,
+};
+
+use crate::cost_model;
+
+// we only accept receipts with minimal 1 wei grt
+const MINIMAL_VALUE: u128 = 1;
+
+pub struct AgoraQuery {
```
anirudh2, line 30: I think it's generally worth adding doc comments to public things.

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+    pub deployment_id: DeploymentId,
+    pub query: String,
+    pub variables: String,
+}
+
+type CostModelMap = Arc<RwLock<HashMap<DeploymentId, CostModel>>>;
+type GlobalModel = Arc<RwLock<Option<CostModel>>>;
+
+pub struct MinimumValue {
```
anirudh2, line 39: Same. Another public thing that I think it's worth documenting.

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+    cost_model_map: CostModelMap,
+    global_model: GlobalModel,
+    watcher_cancel_token: tokio_util::sync::CancellationToken,
+}
+
+struct CostModelWatcher {
+    pgpool: PgPool,
+
```
anirudh2, line 47: Is this newline meant to be here?

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+                Ok(pg_notification) = pglistener.recv() => {
+                    cost_model_watcher.new_notification(
+                        pg_notification,
+                    ).await;
+                }
+            }
+        }
+    }
+
+    async fn new_notification(&self, pg_notification: PgNotification) {
```
anirudh2, line 80: I think this function does too many things. It'd be better if it were broken up. For example,
* `parse_notification`
* `handle_insert`
* `handle_delete`
* `handle_unexpected_notification`

Not necessarily the best breakdown, or naming but just as an illustrative example.

common/src/tap/checks/value_check.rs
```diff
@@ -0,0 +1,535 @@
...
+        ));
+
+        Self {
+            global_model,
+            cost_model_map,
+            watcher_cancel_token,
+        }
+    }
+
+    fn get_expected_value(&self, agora_query: &AgoraQuery) -> anyhow::Result<u128> {
```
anirudh2, line 191: The naming can be somewhat arbitrary. For example, here, you have a getter `get_expected_value`, but in `cost_model.rs`, you have getters named `cost_models` and `cost_model`.

anirudh2: Good with either, but just pick one.

anirudh2 (approved):
Great work, @gusinacio! Lot's of great stuff in here. None of my requests are necessarily blocking, so I'm going to approve the PR. Your call whether to address them or not.

common/src/cost_model.rs
gusinacio: Hey, I was just moving code from `indexer-service` to here to access this code.

common/src/cost_model.rs
gusinacio: Same here.

common/src/tap/checks/value_check.rs
gusinacio: I usually like to separate things inside structs like:
- Things related to database
- Things related to channels
- Things related to config

It's not a split inside the struct itself, but it helps when you want to abstract a whole portion into a separated struct.

common/src/tap/checks/value_check.rs
gusinacio: We have `cost_models` and `cost_model` there because they map to GraphQL queries that indexer-service has.

I usually use `get_` but it's a Java mind hahaha, maybe a future refactor I should remove all getters into more meaningful names.

anirudh2: I love that you thumbs-up'ed your own comment hahaha. Glad you agree with yourself :joy:

common/src/tap/checks/value_check.rs
anirudh2: Much cleaner to me! Thanks!

anirudh2 (approved):
Thanks, @gusinacio!

PR #155, opened 2024-04-26
Add timestamp check for receipts

carlosvdr:


aasseman (commented):
A few suggestions, otherwise logic looks good.
Maybe you could add a unit test? I know it's pretty simple stuff, but the test should be pretty simple too.

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,42 @@
...
+#[async_trait::async_trait]
+impl Check for TimestampCheck {
+    async fn check(&self, receipt: &ReceiptWithState<Checking>) -> CheckResult {
+        let timestamp_now = SystemTime::now().duration_since(SystemTime::UNIX_EPOCH)?;
+        let min_timestamp = timestamp_now - self.timestamp_threshold;
+        let max_timestamp = timestamp_now + self.timestamp_threshold;
+
+        let receipt_timestamp = Duration::from_nanos(receipt.signed_receipt().message.timestamp_ns);
+
+        if receipt_timestamp < max_timestamp && receipt_timestamp > min_timestamp {
```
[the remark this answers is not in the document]
aasseman: Actually nevermind, you can't have negative `Duration`. The solution would be to use https://doc.rust-lang.org/std/time/struct.Duration.html#method.abs_diff but it's experimental.

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,107 @@
...
+    #[tokio::test]
+    async fn test_timestamp_inside_tolerance() {
+        let signed_receipt = create_signed_receipt_with_jitter(15);
+        let timestamp_check = TimestampCheck::new(Duration::from_secs(30));
+        assert!(timestamp_check.check(&signed_receipt).await.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_timestamp_outside_tolerance() {
+        let signed_receipt = create_signed_receipt_with_jitter(31);
```
gusinacio, line 103: This test is testing only if the receipt has + 31 secs. 

Could you create another test where you check if it the timestamp = now - 31 secs?

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,126 @@
...
+    async fn test_timestamp_more_than_tolerance() {
+        let timestamp = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_nanos()
+            + Duration::from_secs(33).as_nanos();
+        let timestamp_ns = timestamp as u64;
+        let signed_receipt = create_signed_receipt_with_jitter(timestamp_ns);
+        let timestamp_check = TimestampCheck::new(Duration::from_secs(30));
+        assert!(timestamp_check.check(&signed_receipt).await.is_err());
```
aasseman, line 111: You can also do
```suggestion
        timestamp_check.check(&signed_receipt).await.unwrap();
```
But you don't have to of course.

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,126 @@
...
+    async fn test_timestamp_inside_tolerance() {
+        let timestamp = SystemTime::now()
+            .duration_since(SystemTime::UNIX_EPOCH)
+            .expect("Time went backwards")
+            .as_nanos()
+            + Duration::from_secs(15).as_nanos();
+        let timestamp_ns = timestamp as u64;
+        let signed_receipt = create_signed_receipt_with_jitter(timestamp_ns);
+        let timestamp_check = TimestampCheck::new(Duration::from_secs(30));
+        assert!(timestamp_check.check(&signed_receipt).await.is_ok());
```
aasseman, line 98: You can also do. It's the same but just FYI.
```suggestion
        timestamp_check.check(&signed_receipt).await.unwrap();
```

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,126 @@
...
+    use tap_core::{
+        receipt::{checks::Check, Checking, Receipt, ReceiptWithState},
+        signed_message::EIP712SignedMessage,
+    };
+
+    use alloy_sol_types::eip712_domain;
+    use alloy_sol_types::Eip712Domain;
+    use ethers::signers::coins_bip39::English;
+    use ethers::signers::{LocalWallet, MnemonicBuilder};
+    fn create_signed_receipt_with_jitter(timestamp_ns: u64) -> ReceiptWithState<Checking> {
```
aasseman, line 57: You removed the built in jitter, so it needs renaming now
```suggestion
    fn create_signed_receipt(timestamp_ns: u64) -> ReceiptWithState<Checking> {
```

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    use tap_core::{
+        receipt::{checks::Check, Checking, Receipt, ReceiptWithState},
+        signed_message::EIP712SignedMessage,
+    };
+
+    use alloy_sol_types::eip712_domain;
+    use alloy_sol_types::Eip712Domain;
+    use ethers::signers::coins_bip39::English;
+    use ethers::signers::{LocalWallet, MnemonicBuilder};
```
aasseman, line 56: Formatting import groups (like in python): stdlib, external deps, internal imports. + Adding a space between imports and first function declaration.
```suggestion
    use std::str::FromStr;

    use alloy_primitives::Address;
    use alloy_sol_types::eip712_domain;
    use alloy_sol_types::Eip712Domain;
    use ethers::signers::coins_bip39::English;
    use ethers::signers::{LocalWallet, MnemonicBuilder};
    
    use super::*;
    use tap_core::{
        receipt::{checks::Check, Checking, Receipt, ReceiptWithState},
        signed_message::EIP712SignedMessage,
    };

```

Only source I found is from the rustfmt docs, they have an automated rule for this but it's still not stable :( https://rust-lang.github.io/rustfmt/?version=v1.7.0&search=grouping#group_imports

aasseman (changes requested):

common/src/tap/checks/timestamp_check.rs
carlosvdr: I do like more visually having the assert since makes it more obvious atleast to me what is going on

common/src/indexer_service/http/config.rs
```diff
@@ -71,4 +72,5 @@ pub struct IndexerConfig {
 pub struct ScalarConfig {
     pub chain_id: u64,
     pub receipts_verifier_address: Address,
+    pub timestamp_error_tolerance: Duration,
```
gusinacio, line 75: @aasseman do you know if it's possible to set default values for this?

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,130 @@
...
+            .as_nanos()
+            + Duration::from_secs(15).as_nanos();
+        let timestamp_ns = timestamp as u64;
+        let signed_receipt = create_signed_receipt_with_custom_timestamp(timestamp_ns);
+        let timestamp_check = TimestampCheck::new(Duration::from_secs(30));
+        assert!(timestamp_check.check(&signed_receipt).await.is_ok());
+    }
+
+    #[tokio::test]
+    async fn test_timestamp_more_than_tolerance() {
```
gusinacio, line 106: Actually, this is `test_timestamp_less_than_tolerance` because the `now` is being modified instead of the timestamp.

common/src/tap/checks/timestamp_check.rs
```diff
@@ -0,0 +1,130 @@
...
+            .as_nanos()
+            + Duration::from_secs(33).as_nanos();
+        let timestamp_ns = timestamp as u64;
+        let signed_receipt = create_signed_receipt_with_custom_timestamp(timestamp_ns);
+        let timestamp_check = TimestampCheck::new(Duration::from_secs(30));
+        assert!(timestamp_check.check(&signed_receipt).await.is_err());
+    }
+
+    #[tokio::test]
+    async fn test_timestamp_less_than_tolerance() {
```
gusinacio, line 119: Same here but inversed.

common/src/indexer_service/http/config.rs
aasseman: The most important thing is that it has to be smaller than `rav_request_timestamp_buffer_ms`

aasseman: I think we should increase the default value of `rav_request_timestamp_buffer_ms` to maybe 30sec or 1min?
Then for ex if we choose 1 min, we could have `timestamp_error_tolerance` default to 30 sec.

aasseman (approved):

aasseman (approved):

PR #159, opened 2024-05-02
Add limit of receipts per rav request

### LINKED ISSUE #96: [Feat.Req] Limit number of receipts per RAV request
**Problem statement**
TAP Agent will currently try to send all unaggregated receipts with a RAV request. 
The TAP Aggregator will currently not accept more than 10MB of data per request (otherwise would be a DoS vector). It was therefore decided that the TAP Aggregator should be guaranteed to handle a maximum of 15,000 receipts per call.

**Expectation proposal**
Implement the logic that will only pull up to 15,000 oldest unaggregated receipts per RAV request, as well as continuing to do more RAV requests if there should be more than 15,000 receipts to aggregate.
This will involve changes in https://github.com/semiotic-ai/timeline-aggregation-protocol/tree/main/tap_core.

**Additional context**
https://github.com/semiotic-ai/timeline-aggregation-protocol/blob/main/tap_aggregator/README.md

carlosvdr:
Added to config a limit  that will define the amount of receipts to aggregate per request
fixes #96

### DIFF
--- a/tap-agent/src/config.rs
+++ b/tap-agent/src/config.rs
@@ -239,7 +239,7 @@ pub struct Tap {
         env = "RAV_REQUEST_TIMESTAMP_BUFFER",
         help = "Buffer (in ms) to add between the current time and the timestamp of the \
         last unaggregated fee when triggering a RAV request.",
-        default_value_t = 1_000 // 1 second
+        default_value_t = 60_000 // 60 seconds
     )]
     pub rav_request_timestamp_buffer_ms: u64,
     #[clap(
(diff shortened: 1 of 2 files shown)


tap-agent/src/config.rs
```diff
@@ -259,6 +259,15 @@ pub struct Tap {
         help = "YAML file with a map of sender addresses to aggregator endpoints."
     )]
     pub sender_aggregator_endpoints_file: PathBuf,
+
+    #[clap(
+        long,
+        value_name = "rav-request-receipt-limit",
+        env = "RAV_REQUEST_RECEIPT_LIMIT",
+        help = "Amount that will define the limit of receipts to aggregate per request",
```
aasseman, line 267:
```suggestion
        help = "Maximum number of receipts per aggregation request",
```

tap-agent/src/config.rs
```diff
@@ -239,7 +239,7 @@ pub struct Tap {
         env = "RAV_REQUEST_TIMESTAMP_BUFFER",
         help = "Buffer (in ms) to add between the current time and the timestamp of the \
         last unaggregated fee when triggering a RAV request.",
-        default_value_t = 1_000 // 1 second
+        default_value_t = 60_000 // 60 seconds
```
aasseman, line 242: This was more relevant for the previous PR than this one, but ok.

carlosvdr: Yes, this was a mistake shouldnt appear since its like that in main already

gusinacio (approved):

Issue #156, opened 2024-04-29
[Feat.Req] Add prometheus metrics to `tap-agent`


aasseman:
To track, for ex.:
- Current unaggregated fees per allocation x sender
- num of aggregations per allocation x sender
- stats on  RAV response time per sender
- total fees per allocation x sender
- other cool stuff?

PR #165, opened 2024-05-13
fix(tap-agent): metrics port config

aasseman:
A metrics config port was unnecessarily added in #156 , and created a conflict in clap that breaks argument parsing completely.


gusinacio:
rust fmt is failing

carlosvdr:
Didn't realized there was already one

carlosvdr (approved):
LGTM

PR #168, opened 2024-05-14
feat(tap-agent): Deny senders over max_unnaggregated_fees_per_sender

### LINKED ISSUE #103: [Feat.Req] Deny senders over a certain amount of unaggregated fees
**Problem statement**
Implement a maximum unaggregated fees trigger (that is larger than the trigger for RAV request) over which business with a sender is to be denied.

**Expectation proposal**
Adding:
- A new unaggregated fees trigger for sender denial
- A sender deny list in DB
- PG notifications to quickly inform the indexer-service instances of adding/removal from the deny list

aasseman:
Fixes #103. (the remaining half)

Adds a new setting `max_unnaggregated_fees_per_sender` (in GRT).
When a sender's unaggregated fees go over that value, it gets added to the `scalar_tap_denylist`, which tells the indexer-service to refuse service from that sender.
When the unaggregated fees come back under the value, it gets removed from the `scalar_tap_denylist`, which tells the indexer-service to accept service from that sender again.

### DIFF
--- a/tap-agent/src/agent/sender_account.rs
+++ b/tap-agent/src/agent/sender_account.rs
@@ -63,7 +63,7 @@ pub struct State {
     sender: Address,
 
     // Deny reasons
-    deny: bool,
+    denied: bool,
 
     //Eventuals
     escrow_accounts: Eventual<EscrowAccounts>,
@@ -136,18 +136,9 @@ impl State {
         Ok(())
     }
 
-    /// Deny the sender if the total fee is greater than the max-unnaggregated-fees-per-sender.
-    /// Will update [`State::deny`], as well as the denylist table in the database.
-    async fn maybe_deny(&mut self) {
-        if self.sender_fee_tracker.get_total_fee()
-            >= self.config.tap.max_unnaggregated_fees_per_sender.into()
-        {
-            tracing::warn!(
-                total_fee = self.sender_fee_tracker.get_total_fee(),
-                max_value = self.config.tap.max_unnaggregated_fees_per_sender,
-                "Total fee greater than max-unnaggregated-fees-per-sender. Denying sender."
-            );
-            self.deny = true;
+    /// Will update [`State::denied`], as well as the denylist table in the database.
+    async fn add_to_denylist(&mut self) {
+        if !self.denied {
             sqlx::query!(
                 r#"
                     INSERT INTO scalar_tap_denylist (sender_address)
@@ -158,21 +149,13 @@ impl State {
             .execute(&self.pgpool)
             .await
             .expect("Should not fail to insert into denylist");
+            self.denied = true;
         }
     }
 
-    /// Allow the sender if the total fee is less than the max-unnaggregated-fees-per-sender.
-    /// Will update [`State::deny`], as well as the denylist table in the database.
-    async fn maybe_allow(&mut self) {
-        if self.sender_fee_tracker.get_total_fee()
-            < self.config.tap.max_unnaggregated_fees_per_sender.into()
-        {
-            tracing::info!(
-                total_fee = self.sender_fee_tracker.get_total_fee(),
-                max_value = self.config.tap.max_unnaggregated_fees_per_sender,
-                "Total fee fell below max-unnaggregated-fees-per-sender. Allowing sender again."
-            );
-            self.deny = false;
+    /// Will update [`State::denied`], as well as the denylist table in the database.
+    async fn remove_from_denylist(&mut self) {
+        if self.denied {
             sqlx::query!(
                 r#"
                     DELETE FROM scalar_tap_denylist 
@@ -183,6 +166,7 @@ impl State {
             .execute(&self.pgpool)
             .await
             .expect("Should not fail to delete from denylist");
+            self.denied = false;
         }
     }
 }
@@ -228,19 +212,19 @@ impl Actor for SenderAccount {
         let escrow_adapter = EscrowAdapter::new(escrow_accounts.clone(), sender_id);
 
         // Get deny status from the scalar_tap_denylist table
-        let deny = sqlx::query!(
+        let denied = sqlx::query!(
             r#"
                 SELECT EXISTS (
                     SELECT 1
                     FROM scalar_tap_denylist
                     WHERE sender_address = $1
-                ) as deny
+                ) as denied
             "#,
             sender_id.encode_hex::<String>(),
         )
         .fetch_one(&pgpool)
         .await?
-        .deny
+        .denied
         .expect("Deny status cannot be null");
 
         let state = State {
@@ -256,7 +240,7 @@ impl Actor for SenderAccount {
             config,
             pgpool,
             sender: sender_id,
-            deny,
+            denied,
         };
 
         for allocation_id in &allocation_ids {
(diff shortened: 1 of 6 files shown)


tap-agent/src/agent/sender_account.rs
```diff
@@ -129,6 +135,56 @@ impl State {
         self.sender_fee_tracker.update(allocation_id, result.value);
         Ok(())
     }
+
+    /// Deny the sender if the total fee is greater than the max-unnaggregated-fees-per-sender.
+    /// Will update [`State::deny`], as well as the denylist table in the database.
+    async fn maybe_deny(&mut self) {
+        if self.sender_fee_tracker.get_total_fee()
+            >= self.config.tap.max_unnaggregated_fees_per_sender.into()
+        {
```
gusinacio, line 144: I'd rather use `add_to_denylist()` and `remove_from_denylist()` and use the condition where it's supposed to be used, we could also have a `should_deny() -> bool` and just update the condition. 

An alternative could also be something like `verify_and_update_denylist()` where the function could add to the denylist or remove depending on the conditions.

tap-agent/src/agent/sender_account.rs
```diff
@@ -59,6 +62,9 @@ pub struct State {
     _indexer_allocations_handle: PipeHandle,
     sender: Address,
 
+    // Deny reasons
+    deny: bool,
```
gusinacio, line 66: I prefer to have this as `sender_denied` or `in_denylist` is more explicative.

tap-agent/src/agent/sender_account.rs
```diff
@@ -214,17 +287,32 @@ impl Actor for SenderAccount {
...
+                state
+                    .sender_fee_tracker
+                    .update(allocation_id, unaggregated_fees.value);
 
-                if tracker.get_total_fee() >= state.config.tap.rav_request_trigger_value.into() {
+                // Eagerly deny the sender (if needed), before the RAV request. To be sure not to
+                // delay the denial because of the RAV request, which could take some time.
+                if !state.deny {
+                    state.maybe_deny().await;
+                }
```
gusinacio, line 298: If this is already a `maybe_deny`, why do you have an `if` stmt.

If the reason is to not call an async function, something like the following makes more sense: 
```
if state.should_deny() {
    state.add_to_denylist().await;
}
```

tap-agent/src/agent/sender_account.rs
```diff
@@ -214,17 +287,32 @@ impl Actor for SenderAccount {
...
                         "Total fee greater than the trigger value. Triggering RAV request"
                     );
                     state.rav_requester_single().await?;
                 }
+
+                // Maybe allow the sender right after the potential RAV request. This way, the
+                // sender can be allowed again as soon as possible if the RAV was successful.
+                if state.deny {
+                    state.maybe_allow().await;
+                }
```
gusinacio, line 315: Same here, could be:
```
if state.should_allow() {
    state.remove_from_denylist().await;
}
```

tap-agent/src/agent/sender_account.rs
```diff
@@ -129,6 +135,56 @@ impl State {
...
+            )
+            .execute(&self.pgpool)
+            .await
+            .expect("Should not fail to insert into denylist");
+        }
+    }
+
+    /// Allow the sender if the total fee is less than the max-unnaggregated-fees-per-sender.
+    /// Will update [`State::deny`], as well as the denylist table in the database.
+    async fn maybe_allow(&mut self) {
```
gusinacio, line 166: `remove_from_denylist`

tap-agent/src/agent/sender_account.rs
```diff
@@ -627,4 +748,109 @@ pub mod tests {
...
+                UnaggregatedReceipts {
+                    value: 1000,
+                    last_id: 11,
+                },
+            ))
+            .unwrap();
+
+        tokio::time::sleep(Duration::from_millis(10)).await;
+
+        // check if sender is denied
```
gusinacio, line 832: Could you update this test to be around the domain limits?

It is important to keep behavior when: 
- max_unaggregated_fees_per_sender = unaggregated_fees - 1
- max_unaggregated_fees_per_sender = unaggregated_fees
- max_unaggregated_fees_per_sender = unaggregated_fees + 1

gusinacio (changes requested):

tap-agent/src/agent/sender_account.rs
aasseman: > An alternative could also be something like verify_and_update_denylist() where the function could add to the denylist or remove depending on the conditions.

I prefer add to denylist before the RAV request, and remove after the RAV request, to remove any opportunity for circumvention (as explained in comments)

tap-agent/src/agent/sender_account.rs
aasseman: The reason was to not spam the DB if the sender is already denied. (and the opposite for the other one). Though I could simplify this further. I think I can simplify this now

tap-agent/src/agent/sender_account.rs
gusinacio: >I prefer add to denylist before the RAV request, and remove after the RAV request, to remove any opportunity for circumvention (as explained in comments)

In that case, you could call it twice 😁 but it's just an alternative.

tap-agent/src/agent/sender_account.rs
gusinacio: Yup, but you could add the `state.deny` inside the `maybe` function

tap-agent/src/agent/sender_account.rs
```diff
@@ -214,17 +287,32 @@ impl Actor for SenderAccount {
...
+                if state.sender_fee_tracker.get_total_fee()
+                    >= state.config.tap.rav_request_trigger_value.into()
+                {
                     tracing::debug!(
-                        total_fee = tracker.get_total_fee(),
+                        total_fee = state.sender_fee_tracker.get_total_fee(),
                         trigger_value = state.config.tap.rav_request_trigger_value,
                         "Total fee greater than the trigger value. Triggering RAV request"
                     );
                     state.rav_requester_single().await?;
```
gusinacio, line 308: You probably need to handle this, otherwise when the `rav_requester_single()` fails, the actor will be on a failed state.

tap-agent/src/agent/sender_account.rs
aasseman: Here's a refactor: https://github.com/graphprotocol/indexer-rs/pull/168/commits/f66879adfeacd7b283203bd5e4e55f283d5ad64d
Let me know if that's what you had in mind

tap-agent/src/agent/sender_account.rs
aasseman: If I get a receipt that pushes me over the limit, but request a RAV instead of immediately denying, I'm exposing myself to more risk. The RAV request can take a lot of time during which I'll continue receiving queries.

tap-agent/src/agent/sender_account.rs
aasseman: That's not code that I changed here though. Perhaps for another PR?

tap-agent/src/agent/sender_account.rs
aasseman: Addressed in https://github.com/graphprotocol/indexer-rs/pull/168/commits/7d79082b764a578332e0430a8a2e97eb8fa0ed25

tap-agent/src/agent/sender_account.rs
gusinacio: sure

gusinacio (approved):
LGTM

PR #169, opened 2024-05-15
fix(tap-agent): pending receipts SQL query performance

### LINKED ISSUE #167: (tap-agent) init sender-allocation SQL query takes too long
This query
https://github.com/graphprotocol/indexer-rs/blob/3dcd0f5774cacf8c94f909eac35c286a76152446/tap-agent/src/agent/sender_accounts_manager.rs#L309-L321
Can take a very long time when there's a lot of receipts in store. Triggers a timeout panic.

Need to do some perf tests to figure out the best course of action.

aasseman:
Fixes #167

Tested locally with 10k receipts. `get_pending_sender_allocation_id()` went from ~13 seconds to ~3ms.
All the receipts were with the same (sender, allocation) pair, but I think it's still close enough to reality where we'd expect not that many different pairs even for many queries.


gusinacio (approved):
LGTM

PR #173, opened 2024-05-15
feat(tap-agent): monitor escrow account balance and allow/deny sender

gusinacio:


tap-agent/src/agent/sender_account.rs
```diff
@@ -130,44 +145,76 @@ impl State {
...
     /// Will update [`State::denied`], as well as the denylist table in the database.
     async fn add_to_denylist(&mut self) {
-        if !self.denied {
-            sqlx::query!(
-                r#"
+        tracing::warn!(
+            total_fee = self.sender_fee_tracker.get_total_fee(),
+            max_value = self.config.tap.max_unnaggregated_fees_per_sender,
+            "Total fee greater than max-unnaggregated-fees-per-sender. Denying sender."
+        );
```
gusinacio, line 184: update the message for the other condition

tap-agent/src/agent/sender_account.rs
```diff
@@ -130,44 +145,76 @@ impl State {
...
-        if self.denied {
-            sqlx::query!(
-                r#"
-                    DELETE FROM scalar_tap_denylist 
+        tracing::info!(
+            total_fee = self.sender_fee_tracker.get_total_fee(),
+            max_value = self.config.tap.max_unnaggregated_fees_per_sender,
+            "Total fee fell below max-unnaggregated-fees-per-sender. Allowing sender \
+            again."
+        );
```
gusinacio, line 206: update message for the other condition

tap-agent/src/agent/sender_account.rs
```diff
@@ -209,6 +256,83 @@ impl Actor for SenderAccount {
...
+            #[derive(Deserialize)]
+            struct Response {
+                transactions: Vec<Transaction>
+            }
+
+            async move {
+                let last_non_final_ravs =
+                    sqlx::query!(
+                        r#"
+                            SELECT allocation_id, value_aggregate FROM scalar_tap_ravs WHERE sender_address = $1 AND last AND NOT final;
```
aasseman, line 285: Super long strings like this break cargo-fmt completely on the whole file.
Gotta make sure string / comments don't push the line over 100 characters.
That's why the file is actually not formatted...
```suggestion
                            SELECT allocation_id, value_aggregate 
                            FROM scalar_tap_ravs 
                            WHERE sender_address = $1 AND last AND NOT final;
```

tap-agent/src/agent/sender_account.rs
```diff
@@ -209,6 +256,83 @@ impl Actor for SenderAccount {
...
+                .expect("Should not fail to fetch from scalar_tap_ravs");
+
+
+                let query = r#"query transactions($unfinalizedRavsAllocationIds: [String!]!, $sender: String!) {
+                              transactions(
+                                where: { type: "redeem", allocationID_in: $unfinalizedRavsAllocationIds, sender_: { id: $sender } }
+                              ) {
+                                allocationID
+                              }
+                            }"#;
```
aasseman, line 300: Same here
```suggestion
                let query = r#"
                    query transactions(
                        $unfinalizedRavsAllocationIds: [String!]!
                        $sender: String!
                    ) {
                        transactions(
                            where: {
                                type: "redeem"
                                allocationID_in: $unfinalizedRavsAllocationIds
                                sender_: { id: $sender }
                            }
                        ) {
                            allocationID
                        }
                    }
                "#;
```

tap-agent/src/agent/sender_account.rs
```diff
@@ -270,6 +404,15 @@ impl Actor for SenderAccount {
...
         );
         match message {
+            SenderAccountMessage::UpdateRav(rav) => {
+                state
+                    .rav_tracker
+                    .update(rav.message.allocationId, rav.message.valueAggregate);
+                let should_deny = !state.denied && state.deny_condition_reached();
+                if should_deny {
+                    state.add_to_denylist().await;
+                }
```
aasseman, line 414: Shouldn't it also have a `remove_from_denylist` here?

gusinacio: ravs can only go up, only unaggregated fees and the other check can go down

tap-agent/src/agent/sender_account.rs
aasseman, line 1: This commit has a bunch of diffs that are not `test`.

aasseman (approved):

PR #174, opened 2024-05-20
fix(tap-agent): ignore last ravs while intializing

### LINKED ISSUE #170: Ignore last ravs during initialization on `get_pending_allocation_ids()`
tap-agent is not filtering for pending ravs

gusinacio:
Closes #170


aasseman:
Thinking in the context of #173, shouldn't we ignore only the `final` ravs? Because `last` ravs that are not `final` should still be taken into account for the check introduced in #173.

gusinacio:
#173 doesn't need an actor spawned to keep track of the Rav because it uses a database query where it looks for `last and not final`.

aasseman (approved):

Issue #128, opened 2024-02-29
Bug: indexer_allocations limited to first 1000


hopeyen:
**Describe the bug**

https://github.com/graphprotocol/indexer-rs/blob/main/common/src/allocations/monitor.rs#L68
This query string limits indexer_allocation query to first 1000 active allocations with no pagination. Shown by an instance that there exists indexer with 1000+ active allocations, this function fails to grab all eligible allocations for monitoring. 

**Expected behavior**
Add pagination to the query and ensure all active allocations for an indexer are monitored.

**Additional context**
Uncovered by the same issue on TS version of indexer-service

aasseman:
Corresponding fix in the TS version: https://github.com/graphprotocol/indexer/pull/862

PR #178, opened 2024-05-20
fix(common): pagination for allocation monitor gql queries

carlosvdr:
Related to #128


aasseman (changes requested):
The tests aren't passing in CI:
```
test allocations::monitor::test::test_parses_allocation_data_from_network_subgraph_correctly has been running for over 60 seconds
```

common/src/allocations/monitor.rs
```diff
@@ -135,8 +46,62 @@ pub fn indexer_allocations(
...
+                subgraphDeployment {{
+                    id
+                    deniedAt
+                }}
+            }}
+        "#,
+        indexer_address.to_string().to_ascii_lowercase(),
+        closed_at_threshold.as_secs(),
+    );
+    //println!("{}",query);
```
gusinacio, line 90: Remove this comment

tap-agent/src/tap/context/checks/allocation_id.rs
```diff
@@ -149,7 +149,7 @@ async fn query_escrow_check_transactions(
         ))
         .await?;
 
-    Ok(response.map(|data| !data.transactions.is_empty())?)
+    Ok(response.map(|data| !data.transactions.is_empty()).unwrap())
```
gusinacio, line 152: Why unwrap it here?

common/src/subgraph_client/client.rs
```diff
@@ -189,7 +229,7 @@ impl SubgraphClient {
     pub async fn query<T: for<'de> Deserialize<'de>>(
         &self,
         query: impl IntoRequestParameters + Send + Clone,
-    ) -> Result<ResponseResult<T>, anyhow::Error> {
```
gusinacio, line 192: Update `ResponseResult` to be `Result<T, String>` on the definition

common/src/allocations/monitor.rs
```diff
@@ -135,8 +46,62 @@ pub fn indexer_allocations(
...
+                allocatedTokens
+                createdAtBlockHash
+                createdAtEpoch
+                closedAtEpoch
+                subgraphDeployment {{
+                    id
+                    deniedAt
+                }}
+            }}
+        "#,
```
gusinacio, line 86: Much better query, great job!

gusinacio (changes requested):

common/src/subgraph_client/client.rs
carlosvdr: Perhaps you saw the old commit?
https://github.com/graphprotocol/indexer-rs/blob/a575b00262c03a34fb92759147fcac0896356892/common/src/subgraph_client/client.rs#L232

aasseman (changes requested):

common/src/subgraph_client/client.rs
gusinacio: nvm, I thought that `ResponseResult` was from our crate, but it's from thegraph-core

common/src/allocations/monitor.rs
```diff
@@ -183,16 +147,41 @@ mod test {
...
+            Duration::from_secs(1712448507),
+        )
+        .await;
+        match result {
+            Ok(allocations) => {
+                assert!(allocations.len() > 1)
+            }
+            Err(e) => println!("Error: {:#?}", e),
+        }
     }
```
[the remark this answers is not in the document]
gusinacio: Speaking of which, in case there's an error it's not failing.

common/src/allocations/monitor.rs
```diff
@@ -148,17 +147,18 @@ mod test {
...
-            *test_vectors::INDEXER_ALLOCATIONS
-        );
+        )
+        .await;
+        match result {
+            Ok(allocations) => {
+                assert_eq!(allocations, *test_vectors::INDEXER_ALLOCATIONS)
+            }
+            Err(e) => println!("Error: {:#?}", e),
+        }
```
gusinacio, line 161: This test won't fail in case of an error. Try to use `assert`s and `panic` instead of println

common/src/allocations/monitor.rs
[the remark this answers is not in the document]
carlosvdr: Even if they are closed they should still appear.. or is there a system that deletes them? 
But yeah depending on external source sucks but not sure how else to really test this is doing what its intended without mocking everything which then it would lose all meaning

common/src/allocations/monitor.rs
```diff
@@ -135,8 +46,61 @@ pub fn indexer_allocations(
...
+                    deniedAt
+                }}
+            }}
+        "#,
+        indexer_address.to_string().to_ascii_lowercase(),
+        closed_at_threshold.as_secs(),
+    );
+
+    let responses = network_subgraph
+        .paginated_query::<Allocation>(query, 1000)
```
aasseman, line 92: Maybe using 200 would be more reasonable. It's the value that's used in the gateway code btw.

common/src/allocations/monitor.rs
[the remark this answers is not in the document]
aasseman: @carlosvdr oh yeah you're right, what makes it guaranteed to return stuff is the `1712448507` timestamp. In that case you can improve that test in 2 ways:
- Make the pages shorter, 1000 -> 200. Not really a test improvement, but I think it's a more reasonable default value.
- Find an indexer / timestamp pair that yields much more than 200 allocations. Since the number of allocations can only increase over time (because the timestamp is hardcoded), you can test like `assert!(allocations.len() > 1234)`

common/src/allocations/monitor.rs
```diff
@@ -173,15 +123,15 @@ mod test {
...
         let result = get_allocations(
             network_subgraph_client(),
-            Address::from_str("0xa7a068c169d3f38efc62a1d6ff8de8b1ddfc8a57").unwrap(),
+            Address::from_str("0x326c584e0f0eab1f1f83c93cc6ae1acc0feba0bc").unwrap(),
             Duration::from_secs(1712448507),
         )
         .await;
         match result {
             Ok(allocations) => {
                 assert!(allocations.len() > 1)
```
aasseman, line 132: How many allocations do we expect here?

aasseman (changes requested):

common/src/allocations/monitor.rs
```diff
@@ -135,64 +45,93 @@ pub fn indexer_allocations(
...
+            Address::from_str("0x326c584e0f0eab1f1f83c93cc6ae1acc0feba0bc").unwrap(),
+            Duration::from_secs(1712448507),
+        )
+        .await;
+        match result {
+            Ok(allocations) => {
+                assert!(allocations.len() > 1)
+            }
+            Err(e) => panic!("Error: {:#?}", e),
+        }
```
aasseman, line 135: And replace `1` with the min number of allocations we're expecting to get from that indexer.
```suggestion
        assert!(result.unwrap().len() > 1)
```

aasseman (changes requested):

common/src/allocations/monitor.rs
carlosvdr: It has more than 2000, 2049 actually, but I dont think we can just "hardcode" a value, cause this number will increase and also there will be a point where the closedAllocations will be off from the threshold, But i do believe setting a number like more than 2000 should be okay for this case (already tried it out)

aasseman: Sure, it can only increase anyway, so it's good

aasseman: I see [812965b](https://github.com/graphprotocol/indexer-rs/pull/178/commits/812965ba829248e7f8bc4a9a5279ddb02428e540) but how about the code refactor that I suggested?

carlosvdr: Could you point me to it again, I rebased with main since there where some conflicts and cant see it no more

aasseman (approved):
LGTM

aasseman (approved):

PR #180, opened 2024-05-21
feat(tap-agent)!: overhaul config mgmt

aasseman:
Making it look more like the config for service: Using only toml files + use shellexpand. 
Note that I also changed the GRT values to u128, because u64 can only go up to ~20GRT or so.

### DIFF
--- a/tap-agent/src/config.rs
+++ b/tap-agent/src/config.rs
@@ -202,8 +202,6 @@ mod tests {
             }};
         }
 
-        parse!("1").unwrap();
-
         assert_eq!(parse!("1").unwrap(), 1_000_000_000_000_000_000);
         assert_eq!(parse!("1.1").unwrap(), 1_100_000_000_000_000_000);
         assert_eq!(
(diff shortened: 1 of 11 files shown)


tap-agent/src/config.rs
```diff
@@ -343,63 +135,112 @@ impl Cli {
...
-            1
-        );
-        assert!(parse_grt_value_to_nonzero_u64("0").is_err());
-        assert!(parse_grt_value_to_nonzero_u64("-1").is_err());
-        assert_eq!(
-            parse_grt_value_to_nonzero_u64("1.0000000000000000001").unwrap(),
+            parse!("1.0000000000000000001").unwrap(),
             1_000_000_000_000_000_000
         );
     }
```
gusinacio, line 220: I like that you are testing here multiple values to check if the deserialization is good. But I'd like to also test the error messages so we are sure that our user will understand and fix the config issue.

Probably something on those lines:
```
assert_eq!(err.to_string(), "GRT value must be greater than 0");
assert_eq!(err.to_string(), "GRT value cannot be represented as a u128 GRT wei value");
```

gusinacio (commented):
Everything seems to be really nice and straightforward, I just remembered some weird error messages related to deserialization of GRT that we could improve and test it.

tap-agent/src/config.rs
aasseman: Indeed, good idea!

aasseman: Addressed in [fe728de](https://github.com/graphprotocol/indexer-rs/pull/180/commits/fe728de6b7d5d008006d3bd07b71b6f3bcc0abd3)

gusinacio (approved):
LGTM

PR #182, opened 2024-05-22
refactor: restructure startup for cleaner main.rs

suchapalaver:
As I was getting myself more familiar with the `service` code, these were some minor changes I ended up making for myself to make the startup and module organization easier to read.  

Signed-off-by: Joseph Livesey <joseph@semiotic.ai>


service/src/main.rs
```diff
@@ -1,229 +1,14 @@
...
-            .route("/cost", post(routes::cost::cost))
-            .route("/status", post(routes::status))
-            .with_state(state),
-    })
-    .await
+async fn main() {
+    if let Err(e) = run().await {
+        eprintln!("Indexer service error: {e}");
+        // https://doc.rust-lang.org/std/process/fn.exit.html
+        drop(e);
```
aasseman, line 11: Why the explicit drop? The program is going to be terminated anyway.

service/src/main.rs
```diff
@@ -1,229 +1,14 @@
...
-        service_impl: SubgraphService::new(state.clone()),
-        extra_routes: Router::new()
-            .route("/cost", post(routes::cost::cost))
-            .route("/status", post(routes::status))
-            .with_state(state),
-    })
-    .await
+async fn main() {
+    if let Err(e) = run().await {
+        eprintln!("Indexer service error: {e}");
```
aasseman, line 9: Why not move the `tracing_subscriber` initialization here, and use `tracing:error` instead?

aasseman (commented):
Good idea! Was long overdue I guess :)

service/src/main.rs
suchapalaver: The idea comes from the docs that I've linked to there, namely:

>If a clean shutdown is needed it is recommended to only call this function at a known point where there are no more destructors left to run.

service/src/main.rs
suchapalaver: This is also inspired by the documentation on `std::process::exit` that I've quoted in response to your previous comment. If we reach this error path everything by this point has been dropped out of scope except for the error that's bubbled up. Interested to hear what you think!

service/src/main.rs
aasseman: I see. That lead me to some googling and I found a solution that I think is even more elegant here considering the post-refactor state of things: https://doc.rust-lang.org/std/process/struct.ExitCode.html

aasseman: Well, maybe I find it more elegant because it looks more like what we'd do in C, and C feels cozy to me :stuck_out_tongue:

service/src/main.rs
aasseman: If we move `tracing_subscriber::fmt::init();` at the start of main, and use `ExitCode` + `return` instead of calling for early termination, we should avoid all of those potential pitfalls, I think.

aasseman (approved):

PR #185, opened 2024-05-22
refactor(database): streamline model merging

suchapalaver:
Another small thing I caught while familiarizing myself with the code, this removes the need to create an intermediate vector when merging models in `database`.

### DIFF
--- a/service/src/database.rs
+++ b/service/src/database.rs
@@ -122,11 +122,7 @@ pub async fn cost_models(
         models = models
             .into_iter()
             .map(|model| merge_global(model, &global_model))
-            .collect();
-
-        // Inject a cost model for all deployments that don't have one
-        models = models
-            .into_iter()
+            // Inject a cost model for all deployments that don't have one
             .chain(
                 deployments_without_models
                     .into_iter()


aasseman:
:thinking: I see the streamlining of merging the 2 bits of code in the global model case into 1. But I don't think the early return actually makes the code mode readable

suchapalaver:
> 🤔 I see the streamlining of merging the 2 bits of code in the global model case into 1. But I don't think the early return actually makes the code mode readable

NBD, I guess it's personal; the need to mutate the variable in the original threw me when I first read it. But the intention here is to remove the need to `collect` twice - if it would matter in this case(?).

I wonder if making things `if let Some(x)` ... `else` would be better since, as you point out, the `return` isn't really early at all.

aasseman:
Yeah the double collect was a bit nasty

aasseman:
Well merging this whole thing

https://github.com/graphprotocol/indexer-rs/blob/edd1a89a5646e01c8eee94c55b94d748948f1ef4/service/src/database.rs#L122-L139

into 1 was nice though

suchapalaver:
> Well merging this whole thing
> 
> https://github.com/graphprotocol/indexer-rs/blob/edd1a89a5646e01c8eee94c55b94d748948f1ef4/service/src/database.rs#L122-L139
> 
> into 1 was nice though

You got it!

aasseman (approved):

PR #188, opened 2024-05-23
Gusinacio/watch after created logs

gusinacio:


tap-agent/src/agent/sender_accounts_manager.rs
```diff
@@ -109,11 +109,6 @@ impl Actor for SenderAccountsManager {
                 'scalar_tap_receipt_notification'",
             );
         // Start the new_receipts_watcher task that will consume from the `pglistener`
```
aasseman, line 111: This comment should be moved too

tap-agent/src/agent/sender_accounts_manager.rs
```diff
@@ -157,7 +152,14 @@ impl Actor for SenderAccountsManager {
...
         }
 
+        state.new_receipts_watcher_handle = Some(tokio::spawn(new_receipts_watcher(
+            pglistener,
+            escrow_accounts,
+            prefix,
+        )));
+
         tracing::info!("SenderAccountManager created!");
+        // start database monitor
```
aasseman, line 162: I think this comment should be removed. Or maybe I don't understand what it means here.

aasseman (changes requested):
LGTM otherwise

PR #189, opened 2024-05-23
fix(tap-agent): monitoring labels and gauge updates

aasseman:
Noticed some typos as well as potentially confusing docs on the monitoring items.
Also, RAV and unaggregated fees gauges were not updated at the right places.

Sorry for not having caught these in review in the first place :face_with_diagonal_mouth:


carlosvdr (approved):
LGTM

PR #195, opened 2024-05-28
feat: auto retry RAVs if sender denied

### LINKED ISSUE #177: [Feat.Req] Retry rav requests when senders are on a blocklist
**Problem statement**
The current only way to trigger a rav request is by receiving receipts. The problem happens when a sender is on a blacklist because no receipts are accepted, thus the sender will be forever blocked until you restart tap-agent which triggers another rav request if the limit is over.

**Expectation proposal**
Create a scheduler that retries rav requests when the unaggregated receipts exceed the max_limit.

**Alternative considerations**
Verify in case of a rav failure if it should retry within the same scheduler.

gusinacio:
Close #177

### DIFF
--- a/tap-agent/src/agent/sender_account.rs
+++ b/tap-agent/src/agent/sender_account.rs
@@ -5,6 +5,8 @@ use bigdecimal::num_bigint::ToBigInt;
 use bigdecimal::ToPrimitive;
 use std::collections::{HashMap, HashSet};
 use std::str::FromStr;
+use std::time::Duration;
+use tokio::task::JoinHandle;
 
 use alloy_primitives::hex::ToHex;
 use alloy_sol_types::Eip712Domain;
@@ -13,7 +15,7 @@ use ethereum_types::U256;
 use eventuals::{Eventual, EventualExt, PipeHandle};
 use indexer_common::subgraph_client::Query;
 use indexer_common::{escrow_accounts::EscrowAccounts, prelude::SubgraphClient};
-use ractor::{call, Actor, ActorProcessingErr, ActorRef, SupervisionEvent};
+use ractor::{call, Actor, ActorProcessingErr, ActorRef, MessagingErr, SupervisionEvent};
 use serde::Deserialize;
 use sqlx::PgPool;
 use tap_core::rav::SignedRAV;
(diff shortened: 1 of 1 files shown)


tap-agent/src/agent/sender_account.rs
```diff
@@ -484,12 +493,23 @@ impl Actor for SenderAccount {
...
 
-                // Maybe allow the sender right after the potential RAV request. This way, the
-                // sender can be allowed again as soon as possible if the RAV was successful.
-
-                let should_allow = state.denied && !state.deny_condition_reached();
-                if should_allow {
-                    state.remove_from_denylist().await;
+                match (state.denied, state.deny_condition_reached()) {
+                    // Maybe allow the sender right after the potential RAV request. This way, the
+                    // sender can be allowed again as soon as possible if the RAV was successful.
```
aasseman, line 498: If this comment is about the `(true, false)` case, then it *will* allow the sender.
```suggestion
                    // Allow the sender right after the potential RAV request. This way, the
                    // sender can be allowed again as soon as possible if the RAV was successful.
```

tap-agent/src/agent/sender_account.rs
```diff
@@ -484,12 +493,23 @@ impl Actor for SenderAccount {
...
+                    (true, false) => state.remove_from_denylist().await,
+                    // if couldn't remove from denylist, resend the message in 30 seconds
+                    // this will trigger another rav request
```
aasseman, line 501: Won't necessarily trigger another RAV request. Depends how much unaggregated fees we have. 
For ex. if the sender is denied because of low escrow and not unaggregated fees, the next call to `UpdateReceiptFees` will not a trigger a RAV.
```suggestion
                    // if couldn't remove from denylist, resend the message in 30 seconds
                    // this may trigger another rav request
```

tap-agent/src/agent/sender_account.rs
```diff
@@ -455,6 +459,11 @@ impl Actor for SenderAccount {
                 }
             }
             SenderAccountMessage::UpdateReceiptFees(allocation_id, unaggregated_fees) => {
+                // new update receipt fee, abort any scheduled rav request
```
aasseman, line 462:
```suggestion
                // If we're here because of a new receipt, abort any scheduled UpdateReceiptFees
```

aasseman (changes requested):
Proposing improvements to the comments

aasseman (approved):

PR #196, opened 2024-05-29
feat(tap-agent): add invalid receipts tracker for deny condition

gusinacio:


tap-agent/src/agent/sender_account.rs
```diff
@@ -458,6 +463,17 @@ impl Actor for SenderAccount {
...
+            SenderAccountMessage::UpdateInvalidReceiptFees(allocation_id, unaggregated_fees) => {
+                state
+                    .invalid_receipts_tracker
+                    .update(allocation_id, unaggregated_fees.value);
+
+                // invalid receipts can't go down
+                let should_deny = !state.denied && state.deny_condition_reached();
+                if should_deny {
+                    state.add_to_denylist().await;
+                }
```
aasseman, line 475: :thinking: yes though I guess at some point we should have a process to let the indexer clear the failed receipts

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -599,8 +652,14 @@ impl SenderAllocationState {
             )
             .execute(&self.pgpool)
             .await
-            .map_err(|e| anyhow!("Failed to store failed receipt: {:?}", e))?;
+            .map_err(|e| anyhow!("Failed to store invalid receipt: {:?}", e))?;
         }
+        self.invalid_receipts_fees = self.calculate_invalid_receipts_fee().await?;
```
aasseman, line 657: Why doing `calculate_invalid_receipts_fee` here? It's already done in `pre_start`.

gusinacio: This function is `store_invalid_receipts`, it's called when a successful rav has a few failed receipts.

We need to update the invalid receipt fees in this case, like what we do with unaggregated fees after every rav.

tap-agent/src/agent/sender_account.rs
gusinacio: Agreed, but this could be done manually by deleting the invalid receipts and restarting tap-agent.

In another case, we could have indexer-cli to communicate with tap-agent and clear the receipts. 

Right now, the only way to update the invalid receipts is by sending a rav request and that rav request contains invalid receipts. We won't have any rav requests if the invalid receipt is the deny cause of a sender.

tap-agent/src/agent/sender_allocation.rs
aasseman: Yeah but the reason we call `calculate_unaggregated_fee` when doing a rav request is that a successful rav request will delete a bunch of receipt from the DB. So we "need"* to resync with what's in the DB after that.

However there is no auto-deletion of receipts from the `scalar_tap_receipts_invalid` table anywhere, so there isn't a need to resync. It effectively can only grow, as pointed out here: https://github.com/graphprotocol/indexer-rs/pull/196/files/5e413c79dbca46edeef34645143938e324762563#diff-58ccff38a18daf85930e0a6a1a0a29557997648b983157e61c418df0f6a3cb1cR471. 
So just adding to `state.invalid_receipts_fees` after having loaded from the DB once in `pre_start` should be sufficient if I understand things correctly.

Notes:
\* well, we don't actually need to resync, we could instead subtract the successful RAV amount from unaggregated_fees. But resyncing makes sure we don't lose track of things accidentally.

gusinacio: Do you think the suggestion is to add to the counter instead of recalculating it?

aasseman: Yup that was what I was thinking about.
However there may be something I don't see here, I may be wrong :shrug:

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -565,7 +615,10 @@ impl SenderAllocationState {
         }
     }
 
-    async fn store_invalid_receipts(&self, receipts: &[ReceiptWithState<Failed>]) -> Result<()> {
+    async fn store_invalid_receipts(
+        &mut self,
```
aasseman, line 619: After the last commit it doesn't need to `mut` anymore
```suggestion
        &self,
```

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -427,7 +476,7 @@ impl SenderAllocationState {
 
     /// Request a RAV from the sender's TAP aggregator. Only one RAV request will be running at a
     /// time through the use of an internal guard.
-    async fn rav_requester_single(&self) -> Result<SignedRAV> {
+    async fn rav_requester_single(&mut self) -> Result<SignedRAV> {
```
aasseman, line 479: Same here
```suggestion
    async fn rav_requester_single(&self) -> Result<SignedRAV> {
```

aasseman (changes requested):

tap-agent/src/agent/sender_allocation.rs
gusinacio: It does because we are updating the self.invalid_receipt_fees.value which is mut.

tap-agent/src/agent/sender_allocation.rs
gusinacio: same here

tap-agent/src/agent/sender_allocation.rs
aasseman: > It does because we are updating the self.invalid_receipt_fees.value which is mut.
Ohh yeah you're right. Though the thing is that `checked_add` does not add in place. That's why right now you don't need mut (I just tested locally).

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -599,8 +652,31 @@ impl SenderAllocationState {
...
+            .map_err(|e| anyhow!("Failed to store invalid receipt: {:?}", e))?;
         }
+        let fees = receipts
+            .iter()
+            .map(|receipt| receipt.signed_receipt().message.value)
+            .sum();
+
+        self.invalid_receipts_fees
+            .value
+            .checked_add(fees)
```
aasseman, line 664:
```suggestion
        self.invalid_receipts_fees.value = self.invalid_receipts_fees
            .value
            .checked_add(fees)
```

aasseman (changes requested):

tap-agent/src/agent/sender_allocation.rs
gusinacio: Thank you so much about this, I also found a bug in the tests and now they are working as expected!

aasseman (approved):

PR #199, opened 2024-06-02
Gusinacio/uniform config

gusinacio:


aasseman (commented):
Could you also do the default configs like in #180 ?

PS: so we don't use `serde_inline_default` in common anymore

config/src/config.rs
```diff
@@ -0,0 +1,292 @@
...
+    // custom validation of the values
+    fn validate(&self) -> Result<(), String> {
+        let one = BigDecimal::from(1);
+        let ten = BigDecimal::from(10);
+        match &self.tap.rav_request.trigger_value_divisor {
+            x if x <= &one => {
+                return Err("trigger_value_divisor must be greater than 1".to_string())
+            }
+            x if x > &one && x < &ten => warn!(
+                "It's recommended that trigger_value_divisor \
```
aasseman, line 74: Wouldn't it be simpler to do this?
```suggestion
            x if *x > 1.into() && *x < 10.into() => warn!(
```

config/src/config.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    // custom validation of the values
+    fn validate(&self) -> Result<(), String> {
+        let one = BigDecimal::from(1);
+        let ten = BigDecimal::from(10);
+        match &self.tap.rav_request.trigger_value_divisor {
+            x if x <= &one => {
+                return Err("trigger_value_divisor must be greater than 1".to_string())
+            }
+            x if x > &one && x < &ten => warn!(
```
aasseman, line 73: Wouldn't it be simpler to do this?
```suggestion
            x if *x > 1.into() && *x < 10.into() => warn!(
```

common/src/attestations/dispute_manager.rs
```diff
@@ -13,7 +13,6 @@ use crate::subgraph_client::{Query, SubgraphClient};
 
 pub fn dispute_manager(
     network_subgraph: &'static SubgraphClient,
-    graph_network_id: u64,
```
aasseman, line 16: I know that we've never encountered a case where this is `!= 1`, but maybe it's useful in certain testing scenarios to be able to change that value. So maybe we should keep it a config?

gusinacio: I mean, the subgraph itself hardcodes as 1. It's the problem that subgraphs don't have global values so they use 1 as the global key. I took a look at it 👀

gusinacio:
> Could you also do the default configs like in #180 ?
> 
> PS: so we don't use `serde_inline_default` in common anymore

We use default_values, but I use `include_str!` to embed it into the compiled binary while keeping it in a separate file.

common/src/attestations/dispute_manager.rs
gusinacio: You can see the code [here](https://github.com/graphprotocol/graph-network-subgraph/blob/036f14b024d5740587a9f3ca19d1b54d1297507d/src/mappings/helpers/helpers.ts#L459)

config/src/config.rs
gusinacio: Thanks for that. I updated that part of the code, but multiplication wasn't possible to do it because of into().

gusinacio:
> PS: so we don't use `serde_inline_default` in common anymore

just removed `serde_inline_default`

common/src/attestations/dispute_manager.rs
aasseman: Ok then

config/default_values.toml
```diff
@@ -15,7 +15,7 @@ host_and_port = "0.0.0.0:7600"
 url_prefix = "/"
 
 [service.tap]
-max_receipt_value = 1000000000000000 # 0.001 GRT
+max_receipt_value_grt = 1000000000000000 # 0.001 GRT
```
aasseman, line 18: Can we use the custom GRT to u128 deserializer here?

config/maximal-config-example.toml
```diff
@@ -93,7 +93,7 @@ serve_escrow_subgraph = false
 # We need this because a large receipt, especially if it's larger than the RAV request trigger,
 # or worse, the unaggregated receipts limit (tap-agent), can cause the indexer to refuse service
 # to the sender for the duration of RAV request timestamp buffer.
-max_receipt_value = 1000000000000000 # 0.001 GRT
+max_receipt_value_grt = 1000000000000000 # 0.001 GRT
```
aasseman, line 96: and here?

config/default_values.toml
gusinacio: Wow, thanks for showing that. Actually we are converting to GRT this is `1000000000000000 GRT`. Let me fix that

aasseman: Oof that'd be a lot of GRT :grimacing:

config/src/grt.rs
```diff
@@ -0,0 +1,77 @@
...
+
+impl<'de> Deserialize<'de> for NonZeroGRT {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        let v = BigDecimal::deserialize(deserializer)?;
+        if v <= 0.into() {
+            return Err(Error::custom("GRT value must be greater than 0"));
+        }
```
aasseman, line 24: Perhaps the "non-zero" check can be moved to the config validation? Though having it here reduces the amount of code for sure.

config/src/config.rs
```diff
@@ -0,0 +1,291 @@
...
+}
+
+#[derive(Debug, Deserialize_repr, Clone)]
+#[cfg_attr(test, derive(PartialEq))]
+#[repr(u64)]
+pub enum TheGraphChainId {
+    Ethereum = 1,
+    Goerli = 5,
+    Arbitrum = 42161,
+    Sepolia = 11155111,
```
aasseman, line 189: Just to keep the Ethereum chains grouped together
```suggestion
    Goerli = 5,
    Sepolia = 11155111,
    Arbitrum = 42161,
```

service/src/config.rs
```diff
@@ -1,69 +1,80 @@
...
+                recently_closed_allocation_buffer_seconds: 0,
+            },
+            graph_network: GraphNetworkConfig {
+                chain_id: value.blockchain.chain_id.clone() as u64,
+            },
+            tap: TapConfig {
+                chain_id: value.blockchain.chain_id as u64,
+                receipts_verifier_address: value.blockchain.receipts_verifier_address,
+                timestamp_error_tolerance: value.tap.rav_request.timestamp_buffer_secs.as_secs(),
+                receipt_max_value: value.service.tap.max_receipt_value_grt.get_value() as u64,
```
aasseman, line 76: GRT values should be u128, otherwise the max we can take is ~20GRT. Not super urgent to fix though with the current GraphQL use cases at current prices.

Perhaps a fix in another PR?

config/src/config.rs
gusinacio: Done

aasseman (approved):

PR #201, opened 2024-06-05
refactor(common,tap-agent): use tap-receipt header

gusinacio:
Continuation of work in https://github.com/edgeandnode/gateway/pull/805


aasseman (approved):
LGTM

PR #202, opened 2024-06-05
chore: release candidate 1.0.0

gusinacio:


aasseman (commented):
Shouldn't we also version `indexer-config` to 1.0.0-rc.1? Or best to wait for 1.0.0 to version it?
And any opinions about `indexer-common` versionning?

gusinacio:
> Shouldn't we also version `indexer-config` to 1.0.0-rc.1? Or best to wait for 1.0.0 to version it? And any opinions about `indexer-common` versionning?

Thanks for pointing out. Gonna update `indexer-config`. For `indexer-common`, I feel that, as a library and not an application, that should be versioned separately considering that other projects than indexer-rs will use it.

I'm not confident enough to release as a version 1.0.0 because I still consider it unstable.

aasseman (approved):

PR #209, opened 2024-06-15
fix(common): Empty response err on GQL return empty list

aasseman:
Temporary fix for the "empty response" error when a GQL query returns (correctly) an empty list.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6513,8 +6513,7 @@ dependencies = [
 [[package]]
 name = "thegraph-core"
 version = "0.4.1"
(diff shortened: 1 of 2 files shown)


gusinacio (approved):
LGTM

carlosvdr (approved):
lgtm

aasseman:
This just got fixed upstream actually: https://github.com/edgeandnode/toolshed/pull/233.
I'm gonna try updating `thegraph-core` to that, and kill this PR if it works.

aasseman:
Better fix in #210

PR #215, opened 2024-06-19
fix(common): denylist rejection msg sender address

### LINKED ISSUE #213: Bug: signer instead of sender address in indexer-service denylisted error
https://github.com/graphprotocol/indexer-rs/blob/6747363bd539ff022d72efdb905c3e720f6dd884/common/src/tap/checks/deny_list_check.rs#L169-L172

aasseman:
Fix #213


gusinacio (approved):
LGTM

PR #218, opened 2024-06-25
fix(tap-agent): unaggregated_fees metric last RAV

aasseman:
The Unaggregated fees metric was not updating on last RAV.


gusinacio (approved):
LGTM

PR #226, opened 2024-06-26
feature: add support for auth-protected subgraph query URLs

Jannis:
This allows to configure a bearer auth token for the network and escrow subgraphs, in case that is required for the endpoint used.

### DIFF
--- a/common/src/subgraph_client/client.rs
+++ b/common/src/subgraph_client/client.rs
@@ -80,6 +80,7 @@ pub struct DeploymentDetails {
     pub deployment: Option<DeploymentId>,
     pub status_url: Option<Url>,
     pub query_url: Url,
+    pub query_auth_token: Option<String>,
 }
 
 impl DeploymentDetails {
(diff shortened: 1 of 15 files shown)


gusinacio:
Hey Jannis, thanks for contributing.

I'm assuming here that this PR is due to the shutdown of the hosted service.

The decentralized network exposes a way to embed the Api Key in the URL, would this still be necessary or is there any other case I'm missing?

Jannis:
> Hey Jannis, thanks for contributing.
> 
> I'm assuming here that this PR is due to the shutdown of the hosted service.
> 
> The decentralized network exposes a way to embed the Api Key in the URL, would this still be necessary or is there any other case I'm missing?

Indexers should be free to use whatever endpoints they want to query the network and escrow subgraphs. This change allows for instance to use endpoints that require an `Authorization: Bearer ...` header. An example would be another indexer with a free query auth token.

common/src/subgraph_client/client.rs
```diff
@@ -93,14 +94,19 @@ impl DeploymentDetails {
...
+        query_url: &str,
+        query_auth_token: Option<String>,
+    ) -> Result<Self, anyhow::Error> {
         Ok(Self {
             deployment: None,
             status_url: None,
             query_url: Url::parse(query_url)?,
+            query_auth_token,
         })
     }
```
gusinacio, line 111: I think it would be better if we have a new function `fn for_query_url_with_token(query_url: &str, query_auth_token: Option<String>)` instead of overwriting the current `for_query_url` function.

service/src/config.rs
```diff
@@ -58,6 +59,7 @@ impl From<MainConfig> for Config {
                 serve_auth_token: value.service.serve_auth_token,
                 deployment: value.subgraphs.escrow.config.deployment_id,
                 query_url: value.subgraphs.escrow.config.query_url.into(),
+                query_auth_token: value.subgraphs.network.config.query_auth_token.clone(),
```
gusinacio, line 62:
```suggestion
                query_auth_token: value.subgraphs.network.config.query_auth_token,
```
The second doesn't need to be cloned.

aasseman (approved):

PR #228, opened 2024-06-26
Fix `query_url` in config

YaroShkvorets:
After recent refactoring, `[subgraphs.*.query_url]` is broken in config resulting in double slash in query url in the client

What happens is `query_url` deserializes into Url, then gets converted to String with a trailing slash (even if originally it didn't have it), then appended with `/subgraphs/id/Qmxxx`, resulting in double-slash that graph-node can't handle
 
i.e. `http://query-node:8000` -> `http://query-node:8000/` -> `http://query-node:8000//subgraphs/id/Qmxxx`

image

This PR fixes it


aasseman (approved):
Thanks! I missed this detail in the logs, and I was looking in the wrong places.

PR #237, opened 2024-06-28
chore(common): disable integration test depending on the hosted service

LNSD:
This integration test depends on the now "defunct" hosted service, so the test always fails.

I am disabling it until we find a solution to this dependency on the hosted service.

### DIFF
--- a/common/src/subgraph_client/client.rs
+++ b/common/src/subgraph_client/client.rs
@@ -355,6 +355,7 @@ mod test {
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


hopeyen (approved):
cool, maybe we can add an api_key secret as a replacement

aasseman:
> cool, maybe we can add an api_key secret as a replacement

We're working on it with the foundation.
@LNSD good idea to disable in the meantime. I was initially hoping to find a solution quicker.

PR #267, opened 2024-07-31
fix(config): warn for unknown config fields

### LINKED ISSUE #259: Config fails if unknown env var with the right prefix exists
This should be removed:
https://github.com/graphprotocol/indexer-rs/blob/dd311d311ffc145edc09d8a931acdff7c61985dc/config/src/config.rs#L58

The issue is that on kubernetes, there are some environment variables that are automatically created with the k8s app name, such as `TAP_AGENT_SERVICE_PORT=7300`.

That makes the tap-agent crash on start because this doesn't correspond to an expect config field.

aasseman:
Fixes #259


aasseman:
First commit is to reproduce the issue, second one is the fix.

PR #268, opened 2024-08-01
refactor(tap-agent): reuse same http client

gusinacio:
We want to reuse the same connection or wait for the older one in case we hit 2 rav requests in a row


aasseman (approved):
LGTM

PR #276, opened 2024-08-02
feat(config): env support for nested fields

aasseman:
Having the possibility of supplying all of the configuration using environment variables is already a requested feature.
This makes sure that we can do so, even for nested configuration fields.

With this, nesting it handled with double underscore in env var name, example:
```
INDEXER_SERVICE_SUBGRAPHS__NETWORK__QUERY_URL = http[...]
```


gusinacio (approved):
LGTM

PR #277, opened 2024-08-02
test(common): ignore tests needing hosted service

aasseman:
Because the hosted service has been shut down.
We don't have a clear solution for an API key yet.


gusinacio (approved):
LGTM

PR #282, opened 2024-08-07
Invalid receipt fix: when all receipts are invalid

### LINKED ISSUE #214: Store invalid receipts even if there are no valid ones in batch
`rav_requester_single` will throw an error if there are no valid receipts to aggregate, before attempting to move the invalid receipts it found to the failed receipts table.

https://github.com/graphprotocol/indexer-rs/blob/6747363bd539ff022d72efdb905c3e720f6dd884/tap-agent/src/agent/sender_allocation.rs#L481-L514

There could be a situation where all the receipts that have an eligible timestamp are invalid, so they don't get to be moved to the invalid receipts table. An extreme situation is where `# invalid receipts >= rav_receipts_limit`, in which case we'd get stuck unable to create a new RAV even if there are more recent valid receipts in store.

carlosvdr:
fixes #214 
Added function along with this fix to [tap-core](https://github.com/semiotic-ai/timeline-aggregation-protocol/pull/232)


carlosvdr:
this PR depends on [this](https://github.com/graphprotocol/indexer-rs/pull/278) to merge
And also on this [tap_core pr](https://github.com/semiotic-ai/timeline-aggregation-protocol/pull/232) so that I can set a normal version of tap_core at the cargo.toml instead of referencing git

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -508,6 +508,45 @@ impl SenderAllocationState {
                 ),
                 _ => e.into(),
             })?;
+
+        // store invalid receipts and if all are invalid remove them
+        if !invalid_receipts.is_empty() {
```
gusinacio, line 513: you need to check if the valid receipts list is empty

carlosvdr: Hmmm I disagree, the logic before was okay in this, checking if there are `invalid_receipts` if so save them.
Now what I added inside of this if is that  if there are also no `valid_receipts` then yeah we remove them (the `invalid_receipts`).

Because if there are `valid_receipts` in there, then the rav creation process will take care of removing them

gusinacio: Ok, I see that you added the `if valid_receipts.is_empty()` on line 525. But there's one problem that the older solution solved:

What if the request to tap-aggregator fails?

With the older solution, nothing would be stored on the database because it only stores it after the network request.

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -476,6 +476,45 @@ impl SenderAllocationState {
...
+                    .message
+                    .timestamp_ns;
+                sqlx::query!(
+                    r#"
+                        DELETE FROM scalar_tap_receipts
+                        WHERE timestamp_ns BETWEEN $1 AND $2;
+                    "#,
+                    BigDecimal::from(min_timestamp),
+                    BigDecimal::from(max_timestamp)
+                )
```
aasseman, line 513: This is not gonna work correctly I think.

Right now this is where obsolete receipts are deleted: https://github.com/graphprotocol/indexer-rs/blob/d8d1af901cc6ce4dbc65303e7c4ae205c41d8dee/tap-agent/src/agent/sender_allocation.rs#L322. The gist of it is that from time to time, we look at the latest RAV we got in the DB (which must be valid), and delete all receipts that are older than that.
In the case where we got a successful RAV and there were a few failed receipts that were not included, it's fine because the failed receipts, even though not aggregated, will be deleted if they're older than the last RAV.

So the fix we really need is only for the special case of the RAV request failing because *none* of the receipts were valid in the batch. In that case we're stuck trying over and over again because there is no newer RAV that will make it so that eventually the batch of bad receipts gets deleted.
So we'd only look for [this error](https://github.com/semiotic-ai/timeline-aggregation-protocol/blob/eb8447ed4566ced6846c03b510b25b915f985186/tap_core/src/manager/tap_manager.rs#L212) or just `valid_receipts` being empty, and handle the special case there.

We have to be careful with the `BETWEEN` to make sure that it is inclusive both on the lower and higher bound (though please double check it's what we indeed need here).

Also we may consider adding an adapter in tap_core for the removal of receipts in a range, just to make more uniform.

aasseman (changes requested):

tap-agent/src/agent/sender_allocation.rs
aasseman: Ok so I was wrong here, I was confused and missed the `if valid_receipts.is_empty()`. The logic is OK but perhaps there is a way to make it easier to see.

carlosvdr: yeah, switched  to use a match using `expected_rav` and added also `invalid` and `valid` receipts into a tuple as sugested

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -466,92 +466,126 @@ impl SenderAllocationState {
...
-            // Save invalid receipts to the database for logs.
-            // TODO: consider doing that in a spawned task?
-            self.store_invalid_receipts(invalid_receipts.as_slice())
+        match (
+            expected_rav,
+            valid_receipts.is_empty(),
+            invalid_receipts.is_empty(),
+        ) {
+            // No valid receipts, but there are invalid
+            (Err(tap_core::Error::NoValidReceiptsForRAVRequest), true, false) => {
```
aasseman, line 476: I think in this particular case we can simply match any error with `Err(_)`

gusinacio: Agreed, the only possible error for the expected_rav is the `NoValidReceiptsForRAVRequest`

carlosvdr: I believe i already changed it to that

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -466,92 +466,132 @@ impl SenderAllocationState {
...
-
-            // Save invalid receipts to the database for logs.
-            // TODO: consider doing that in a spawned task?
-            self.store_invalid_receipts(invalid_receipts.as_slice())
+        match (
+            expected_rav,
+            valid_receipts.is_empty(),
+            invalid_receipts.is_empty(),
+        ) {
+            // No valid receipts, but there are invalid
```
aasseman, line 475:
```suggestion
            // All receipts are invalid
```
More straight to the point

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -466,92 +466,132 @@ impl SenderAllocationState {
...
+                    .expect("There should be atleast one receipt")
+                    .signed_receipt()
+                    .message
+                    .timestamp_ns;
+                let max_timestamp = invalid_receipts
+                    .last()
+                    .expect("There should be atleast one receipt")
+                    .signed_receipt()
+                    .message
+                    .timestamp_ns;
```
aasseman, line 496: Can you add a comment here about the fact that we expect the receipts to already be sorted by timestamp since we're using [`safe_truncate_receipts`](https://github.com/semiotic-ai/timeline-aggregation-protocol/blob/04b134a4e0268f5b3e87c49639e4d8315b59017a/tap_core/src/manager/adapters/receipt.rs#L102) which happens to do the sort for us [here](https://github.com/graphprotocol/indexer-rs/blob/20bae14e5bfb8e963c444c6dcd475ff0332155d1/tap-agent/src/tap/context/receipt.rs#L157).

aasseman: And also [`collect_receipts`](https://github.com/semiotic-ai/timeline-aggregation-protocol/blob/04b134a4e0268f5b3e87c49639e4d8315b59017a/tap_core/src/manager/tap_manager.rs#L100) doesn't change the order of receipts when checking / moving receipts to the invalid receipts list.

tap-agent/src/agent/sender_allocation.rs
aasseman: I don't see it changed.
This is what I meant btw:
```suggestion
            (Err(_), true, false) => {
```

gusinacio: I talked to @carlosvdr and there's a chance that it fails with "Overflow", so it's nice to have this case explicit.

carlosvdr: Yeah aswell it makes the code more explicit that this specific path should **only** happen for that specific error
Any other error just panic
Makes it clearer IMO

aasseman: Ok, fine with me

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -68,6 +69,21 @@ lazy_static! {
...
+#[derive(Error, Debug)]
+pub enum RavRequesterSingleErrors {
+    #[error(transparent)]
+    SqlxError(#[from] sqlx::Error),
+
+    #[error(transparent)]
+    TapCoreError(#[from] tap_core::Error),
+
+    #[error(transparent)]
+    ClientError(#[from] jsonrpsee::core::ClientError),
```
gusinacio, line 81: You can remove the suffix "Error" from all of these.

gusinacio (approved):
LGTM

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -482,35 +505,33 @@ impl SenderAllocationState {
...
                 let max_timestamp = invalid_receipts
-                    .last()
-                    .expect("There should be atleast one receipt")
-                    .signed_receipt()
-                    .message
-                    .timestamp_ns;
+                    .iter()
+                    .map(|receipt| receipt.signed_receipt().message.timestamp_ns)
+                    .max()
+                    .expect("invalid receitps should not be empty");
```
aasseman, line 518:
```suggestion
                    .expect("invalid receipts should not be empty");
```

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -482,35 +505,33 @@ impl SenderAllocationState {
...
                 let min_timestamp = invalid_receipts
-                    .first()
-                    .expect("There should be atleast one receipt")
-                    .signed_receipt()
-                    .message
-                    .timestamp_ns;
+                    .iter()
+                    .map(|receipt| receipt.signed_receipt().message.timestamp_ns)
+                    .min()
+                    .expect("invalid receitps should not be empty");
```
aasseman, line 513:
```suggestion
                    .expect("invalid receipts should not be empty");
```

aasseman (changes requested):
LGTM but 2 typos to fix, and then need to rebase one last time onto latest main

gusinacio (approved):
LGTM

aasseman (approved):

PR #283, opened 2024-08-09
refactor: non-blocking backoff

### LINKED ISSUE #271: Exponential Backoff for Rav Requests
**Describe the bug**
We were receiving 429 on the rav requests and this was making postgres crazy, we could handle a backoff to limit errors.

We already have a backoff that retries 3 times, but we were getting these errors across retries so we need some way to backoff inter rav requests.

gusinacio:
Fixes #271

### DIFF
--- a/tap-agent/src/agent/sender_allocation.rs
+++ b/tap-agent/src/agent/sender_allocation.rs
@@ -87,6 +87,7 @@ pub struct SenderAllocationState {
     sender_account_ref: ActorRef<SenderAccountMessage>,
 
     failed_ravs_count: u32,
+    failed_rav_backoff: Instant,
 
     http_client: jsonrpsee::http_client::HttpClient,
 }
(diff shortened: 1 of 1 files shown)


tap-agent/src/agent/sender_allocation.rs
```diff
@@ -235,8 +238,15 @@ impl Actor for SenderAllocation {
...
                     // auto backoff retry, on error ignore
-                    if let Err(err) = state.request_rav().await {
-                        error!(error = %err, "Error while requesting rav.");
+                    if Instant::now() > state.failed_rav_backoff {
+                        if let Err(err) = state.request_rav().await {
+                            error!(error = %err, "Error while requesting rav.");
+                        }
+                    } else {
+                        error!(
+                            "Cann't trigger rav request until {:?}",
```
aasseman, line 247:
```suggestion
                            "Can't trigger rav request until {:?} (backoff)",
```

aasseman (changes requested):
LGTM
Just a little fix on a log output

tap-agent/src/agent/sender_allocation.rs
gusinacio: Done

aasseman (approved):

PR #285, opened 2024-08-09
feat(common): batch insert receipts via mpsc

gusinacio:


common/src/tap/receipt_store.rs
```diff
@@ -21,33 +104,8 @@ impl ReceiptStore for IndexerTapContext {
...
-            BigDecimal::from(receipt.message.timestamp_ns),
-            BigDecimal::from(receipt.message.nonce),
-            BigDecimal::from(BigInt::from(receipt.message.value)),
-        )
-        .execute(&self.pgpool)
-        .await
-        .map_err(|e| {
+        let db_receipt = DatabaseReceipt::from_receipt(receipt, &self.domain_separator)?;
+        self.receipt_producer.send(db_receipt).await.map_err(|e| {
             error!("Failed to store receipt: {}", e);
```
aasseman, line 109:
```suggestion
            error!("Failed to queue receipt for storage: {}", e);
```

common/src/tap/receipt_store.rs
```diff
@@ -2,17 +2,100 @@
...
+        inner_context: InnerContext,
+        mut receiver: Receiver<DatabaseReceipt>,
+        cancelation_token: CancellationToken,
+    ) -> JoinHandle<()> {
+        const BUFFER_SIZE: usize = 100;
+        tokio::spawn(async move {
+            loop {
+                let mut buffer = Vec::with_capacity(BUFFER_SIZE);
+                select! {
+                    _ = cancelation_token.cancelled() => { break },
```
aasseman, line 87: Can we make sure that all the remaining receipts are flushed on shudown?

common/src/tap/receipt_store.rs
gusinacio: Done

common/src/tap/receipt_store.rs
aasseman: Wow, I had a hard time finding the docs on `biased`.

aasseman (approved):

common/src/tap/receipt_store.rs
```diff
@@ -21,38 +104,53 @@ impl ReceiptStore for IndexerTapContext {
         &self,
         receipt: ReceiptWithState<Checking>,
     ) -> Result<u64, Self::AdapterError> {
+        let db_receipt = DatabaseReceipt::from_receipt(receipt, &self.domain_separator)?;
+        self.receipt_producer.send(db_receipt).await.map_err(|e| {
+            error!("Failed to queue receipt for storage: {}", e);
+            anyhow!(e)
+        })?;
+
+        // We don't need receipt_ids
```
suchapalaver, line 113: The issue is that all errors within that task are unhandled and this isn't visible to the reader now.

common/src/tap/receipt_store.rs
```diff
@@ -1,18 +1,101 @@
...
+        const BUFFER_SIZE: usize = 100;
+        tokio::spawn(async move {
+            loop {
+                let mut buffer = Vec::with_capacity(BUFFER_SIZE);
+                select! {
+                    biased;
+                    _ = receiver.recv_many(&mut buffer, BUFFER_SIZE) => {
+                        if let Err(e) = inner_context.store_receipts(buffer).await {
+                            error!("Failed to store receipts: {}", e);
+                        }
```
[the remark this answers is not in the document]
aasseman: imo this is a matter of taste here. You can:
- Just send an error out to the logs (current solution), and hope that the indexer will notice that something is wrong.
- Block or crash the whole indexer-service. Higher likelihood that the indexer will notice that something is wrong, but it will also impact its ranking wrt. to the Gateway severely.

Best here is to ask indexer how they would like this error to be handled.

common/src/tap/receipt_store.rs
aasseman: Same thing here, ask what the indexers would prefer.
All the business is going through the indexer-service, so whatever is does, it will be very impactful to the indexers' business.

PR #293, opened 2024-08-19
feat(all): add retryable checks

gusinacio:


aasseman (approved):

common/src/tap/checks/sender_balance_check.rs
```diff
@@ -38,22 +38,25 @@ impl Check for SenderBalanceCheck {
...
+            .map_err(|e| CheckError::Failed(e.into()))?;
 
         // Check that the sender has a non-zero balance -- more advanced accounting is done in
         // `tap-agent`.
         if !escrow_accounts_snapshot
             .get_balance_for_sender(&receipt_sender)
             .map_or(false, |balance| balance > U256::ZERO)
         {
-            return Err(anyhow!(
+            return Err(CheckError::Failed(anyhow!(
```
aasseman, line 56: This can be retryable. Throwing away all the receipts until the sender tops off their balance is not helpful. Better wait for it. We'll keep track of the unaggregated fees either way.

aasseman (changes requested):

common/src/tap/checks/sender_balance_check.rs
gusinacio: The only problem is that the check is on the "common" side, not the tap-agent side. We don't accept it if it fails either way, should we?

gusinacio: Also, this specifc test is just to check if we have a balance greater than zero. Most of the time, this will be someone who never deposited. This could open a precedence of people just creating senders without depositing anything and using the max amount willing to lose.

aasseman: Ah yeah you're right, this is the wrong balance check for this

aasseman (approved):
LGTM

PR #300, opened 2024-08-22
fix(tap-agent): deny sender on failure to create

gusinacio:


aasseman (approved):
LGTM, thanks for catching that

PR #85, opened 2023-11-02
chore: add Edge & Node to copyright headers

hopeyen:
`// Copyright 2023-, GraphOps, and Semiotic Labs.`
-> 
`// Copyright 2023-, Edge & Node, GraphOps, and Semiotic Labs.`

PR #307, opened 2024-08-29
chore: update copyright headers

carlosvdr:
closes #85

### DIFF
--- a/.github/workflows/license_headers_check.yml
+++ b/.github/workflows/license_headers_check.yml
@@ -21,7 +21,7 @@ jobs:
         run: >
           addlicense \
             -check \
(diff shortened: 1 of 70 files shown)


aasseman (changes requested):
The related CI check needs to be adjusted too: https://github.com/graphprotocol/indexer-rs/blob/cbbe1f589a38356c34b2577bdf6af13a1055d4c9/.github/workflows/license_headers_check.yml

However, I don't understand why it didn't scream here :thinking:

aasseman (approved):
Thanks for handling this boring task :pray:

gusinacio (approved):

PR #315, opened 2024-09-19
fix: Remove max_willing_to_lose

### LINKED ISSUE #312: Remove `max_willing_to_loose` from minimal config template
I think it's a bit too dangerous to encourage beginners to play with that setting. 
Let's add the default value of `20` to the default values file.

carlosvdr:
Add max_willing_to_lose into the default config values
fixes #312

### DIFF
--- a/config/minimal-config-example.toml
+++ b/config/minimal-config-example.toml
@@ -51,14 +51,6 @@ receipts_verifier_address = "0x2222222222222222222222222222222222222222"
 ########################################
 # Specific configurations to tap-agent #
 ########################################
-[tap]
-# This is the amount of fees you are willing to risk at any given time. For ex.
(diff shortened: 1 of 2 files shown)


config/minimal-config-example.toml
```diff
@@ -59,7 +59,6 @@ receipts_verifier_address = "0x2222222222222222222222222222222222222222"
 # NOTE: Use strings for decimal values to prevent rounding errors
 # e.g:
 # max_amount_willing_to_lose_grt = "0.1"
-max_amount_willing_to_lose_grt = 20
```
gusinacio, line 62: I think you could remove the whole selection from line 54 until 61

gusinacio: we already have this explanation on the maximal configuration

gusinacio (approved):
LGTM

PR #317, opened 2024-09-19
fix: Rename trigger rav to Rav trigger

### LINKED ISSUE #313: Metrics dashboard: wrong value for "trigger rav"
The "trigger rav" (which could be better called "RAV trigger"), is hardcoded to expect a trigger value divisor of 1000.

Either remove "trigger rav", or gather the divisor from the config somehow (through a new metric for example).

carlosvdr:
Change hardcoded trigger value of 1000 to a metric tap_rav_request_trigger_value
fixes #313

### DIFF
--- a/docs/dashboard.json
+++ b/docs/dashboard.json
@@ -231,7 +231,7 @@
             "uid": "b70befc5-0872-448c-b502-9875d467edaf"
           },
           "editorMode": "code",
-          "expr": "tap_max_fee_per_sender_grt_total * 10^-18/tap_rav_request_trigger_value",
+          "expr": "tap_rav_request_trigger_value * 10^-18",
           "hide": false,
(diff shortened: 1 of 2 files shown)


docs/dashboard.json
```diff
@@ -231,9 +231,9 @@
             "uid": "b70befc5-0872-448c-b502-9875d467edaf"
           },
           "editorMode": "code",
-          "expr": "tap_max_fee_per_sender_grt_total * 10^-18/1000",
+          "expr": "tap_max_fee_per_sender_grt_total * 10^-18/tap_rav_request_trigger_value",
```
gusinacio, line 234: This should probably be just: `tap_rav_request_trigger_value * 10^-18`

gusinacio (changes requested):

gusinacio (approved):
LGTM

PR #322, opened 2024-09-20
fix: Pass receipt error into db

### LINKED ISSUE #272: Display more information of failed receipts
**Describe the bug**
When a receipt fails, we don't know the reason. We want to be able to access the errors and print on the console or even log on to the database.

carlosvdr:
fixes #272


gusinacio (approved):
LGTM

PR #324, opened 2024-09-20
feat: add tracker for buffer unaggregated fees

### LINKED ISSUE #289: Track unaggregated fees inside and outside the buffer
**Describe the bug**
We have a buffer in receipts so we can only create a RAV request outside this buffer window. The problem is that this buffer is not being used inside SenderFeeTracker to trigger RAV requests. Sometimes, one allocation has the biggest unaggregated fees, but all of them are inside that buffer so it doesn't have any valid receipts.

We should be able to keep track of the fees outside the buffer so we can trigger rav requests based on it instead of total fees.

A possible alternative to this solution is to trigger the next biggest allocation if there's "No Receipts found" error.

gusinacio:
Closes #289

### DIFF
--- a/tap-agent/src/agent/sender_allocation.rs
+++ b/tap-agent/src/agent/sender_allocation.rs
@@ -249,7 +249,7 @@ impl Actor for SenderAllocation {
                         .sender_account_ref
                         .cast(SenderAccountMessage::UpdateReceiptFees(
                             state.allocation_id,
-                            ReceiptFees::NewReceipt(notification),
+                            ReceiptFees::NewReceipt(fees),
                         ))?;
                 }
             }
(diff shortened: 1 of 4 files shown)


tap-agent/src/agent/sender_allocation.rs
```diff
@@ -225,9 +225,10 @@ impl Actor for SenderAllocation {
         );
         let unaggregated_fees = &mut state.unaggregated_fees;
         match message {
-            SenderAllocationMessage::NewReceipt(NewReceiptNotification {
-                id, value: fees, ..
-            }) => {
+            SenderAllocationMessage::NewReceipt(notification) => {
```
gusinacio, line 228: This update mostly solves the problem but it's not 100% accurate. We have a system where we trigger a complete re-calculation of unnagregated receipts. When this happens, newer receipts that are in the queue would already be on the counter and that's why we check if `id > unaggregated_fees.last_id` so we don't have a double counter on that.

The problem is that current system for buffer needs them to be called via `add` instead of `update` to be added to the buffer counter. That means that information about what is in buffer is lost under higher loads.

For this to be updated accordingly, tracker would need to have an internal last_id meaning that the value is added to the buffer, but not to the total tracker.

This is not a big issue since the `update`  is usually called in rav requests which reduces total amount, but this issue must be tracked in case problems happen.

tap-agent/src/agent/sender_fee_tracker.rs
```diff
@@ -70,22 +136,31 @@ impl SenderFeeTracker {
     pub fn get_total_fee(&self) -> u128 {
         self.total_fee
     }
+
+    pub fn get_total_fee_outsite_buffer(&mut self) -> u128 {
```
carlosvdr, line 140: tyope outsite => outside

carlosvdr (approved):
lgtm

PR #325, opened 2024-09-20
fix: Add more information if rav request is timing out

### LINKED ISSUE #320: Add error log on RAV Request timeout
Users should know when a RAV Request is timing out and how to fix this by updating their configuration.

Probably an `inspect_err` [here](https://github.com/graphprotocol/indexer-rs/blob/c693f0ebe36a0f5dce8f46fd974eef1a5924c3c6/tap-agent/src/agent/sender_allocation.rs#L547) and a match to print this message only when you have a Timeout.

carlosvdr:
fixes #320

### DIFF
--- a/tap-agent/src/agent/sender_allocation.rs
+++ b/tap-agent/src/agent/sender_allocation.rs
@@ -554,7 +554,18 @@ impl SenderAllocationState {
                             previous_rav
                         ),
                     )
-                    .await?;
+                    .await
+                    .inspect_err(|err| match &err {
+                        jsonrpsee::core::ClientError::RequestTimeout => {
+                            warn!(
+                                "Rav request is timing out, maybe request_timeout_secs is too \
+                                    low in your config file, try adding more secs to the value"
+                            );
+                        }
+                        _ => {
+                            warn!("An error occured {:?}", err);
+                        }
+                    })?;
 
                 let rav_response_time = rav_response_time_start.elapsed();
                 RAV_RESPONSE_TIME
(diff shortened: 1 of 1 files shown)


tap-agent/src/agent/sender_allocation.rs
```diff
@@ -554,7 +554,18 @@ impl SenderAllocationState {
...
+                    .await
+                    .inspect_err(|err| match &err {
+                        jsonrpsee::core::ClientError::RequestTimeout => {
+                            warn!(
+                                "Rav request is timing out, maybe request_timeout_secs is too \
+                                    low in your config file, try adding more secs to the value"
+                            );
+                        }
+                        _ => {
+                            warn!("An error occured {:?}", err);
```
gusinacio, line 566: you can remove this warn because it's already print it out as error somewhere else

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -554,7 +554,18 @@ impl SenderAllocationState {
                             previous_rav
                         ),
                     )
-                    .await?;
+                    .await
+                    .inspect_err(|err| match &err {
+                        jsonrpsee::core::ClientError::RequestTimeout => {
+                            warn!(
+                                "Rav request is timing out, maybe request_timeout_secs is too \
+                                    low in your config file, try adding more secs to the value"
```
gusinacio, line 562: Let the user know if they already raised and it's still getting the same problem to create an issue

gusinacio (changes requested):

gusinacio (approved):
LGTM

PR #326, opened 2024-09-24
Bulk insert of failed receipts

### LINKED ISSUE #273: [Feat.Req] Bulk insert of failed receipts
**Problem statement**
We have a loop that inserts all the failed receipts, but it's way more efficient if we insert via bulk.

taslimmuhammed:
Fixes #273 
Sorry for the delay, I have tested the code[query creation  part] by creating a similar table but not in the entire project, if possible please test it on your environment.

### DIFF
--- a/tap-agent/src/agent/sender_allocation.rs
+++ b/tap-agent/src/agent/sender_allocation.rs
@@ -685,31 +685,14 @@ impl SenderAllocationState {
         &mut self,
         receipts: &[ReceiptWithState<Failed>],
     ) -> Result<()> {
-        let mut query_str = String::from(
-            "INSERT INTO scalar_tap_receipts_invalid (
-                        signer_address,
-                        signature,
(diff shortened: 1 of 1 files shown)


gusinacio (commented):
Hello, first of all thanks for the contribution! From my point of view, this all makes sense and it should work properly.

One key point is that this approach misses `sqlx` greatest feature which is compile time checks. This is really important in case someone changes the code in the future so we know that it compiles to a valid sql.

Could you rewrite that using sqlx macros for compile-time checks?

gusinacio:
You can find an example of how to do that using this as an example.

https://github.com/graphprotocol/indexer-rs/blob/5edf6cfa31900fd3b99ff9a7e586501d7a6a281c/common/src/tap/receipt_store.rs#L25

taslimmuhammed:
Thanks for the review, I'll try using macros

PR #328, opened 2024-09-27
 Bulk insert of failed receipts using query macro

### LINKED ISSUE #273: [Feat.Req] Bulk insert of failed receipts

taslimmuhammed:
Fixes #273 
After running the commands and successful compilation:-
cargo clippy --all-targets --all-features 
cargo fmt 
cargo sqlx prepare --workspace -- --all-targets


tap-agent/src/agent/sender_allocation.rs
```diff
@@ -696,31 +704,43 @@ impl SenderAllocationState {
...
+            values.push(BigDecimal::from(BigInt::from(receipt.message.value)));
         }
+        sqlx::query!(
+            r#"INSERT INTO scalar_tap_receipts_invalid (
+                signer_address,
+                signature,
+                allocation_id,
+                timestamp_ns,
+                nonce,
+                value
```
gusinacio, line 721: It's missing here `error_log`

gusinacio (changes requested):
Invalid receipts have an extra field where we are able to store the reason why the receipt failed.

This new implementation is missing it. Let me know if you need help to add them.

tap-agent/src/agent/sender_allocation.rs
taslimmuhammed: did you meant this line 
debug!(
                    "Receipt for allocation {} and signer {} failed reason: {}",
                    allocation_id.encode_hex(),
                    receipt_signer.encode_hex(),
                    receipt_error
            );

gusinacio: We want to store the error logs for all receipts as well.

Take a look at the older version, we were inserting `error_log` in the old sql.

PR #331, opened 2024-09-30
add dips graphql

mangas:


common/src/indexer_service/http/config.rs
```diff
@@ -42,6 +42,13 @@ pub struct IndexerServiceConfig {
     pub escrow_subgraph: SubgraphConfig,
     pub graph_network: GraphNetworkConfig,
     pub tap: TapConfig,
+    pub dips: IndexerDipsConfig,
+}
+
+#[derive(Debug, Deserialize, Serialize, Clone)]
+pub struct IndexerDipsConfig {
+    pub expected_payee: String,
+    pub allowed_payers: Vec<String>,
```
LNSD, line 51: Let's use the `thegraph_core::Address` type here, as we'll provide Ethereum addresses. This way, we can validate at deserialization time.

config/src/config.rs
```diff
@@ -307,6 +308,13 @@ pub struct TapConfig {
     pub sender_aggregator_endpoints: HashMap<Address, Url>,
 }
 
+#[derive(Debug, Deserialize)]
+#[cfg_attr(test, derive(PartialEq))]
+pub struct DipsConfig {
+    pub expected_payee: String,
+    pub allowed_payers: Vec<String>,
+}
```
LNSD, line 316: Same thing here

dips/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+
+    // the vouchers are generic to each data service, in the case of subgraphs this is an ABI-encoded SubgraphIndexingVoucherMetadata
+    #[derive(Debug, RlpEncodable, RlpDecodable, PartialEq)]
+    struct SubgraphIndexingVoucherMetadata {
+        uint256 pricePerBlock; // wei GRT
+        bytes32 protocolNetwork; // eip199:1 format
+        // differentiate based on indexed chain
+        bytes32 chainId; // eip199:1 format
+    }
+}
```
LNSD, line 53: Note for the future: We might want this to be part of the `thegraph_core` crate to share the same code between the different parts.

dips/src/lib.rs
```diff
@@ -0,0 +1,173 @@
...
+        allowed_payers: impl AsRef<[Address]>,
+    ) -> bool {
+        let sig = match Signature::from_str(&self.signature.to_string()) {
+            Ok(s) => s,
+            Err(_) => return false,
+        };
+
+        let payer = sig
+            .recover_address_from_msg(self.voucher.eip712_hash_struct())
+            .unwrap();
```
LNSD, line 69: I prefer to return a `false` instead of panicking.

EDIT: +1 to the error type on the TODO comment.

dips/src/lib.rs
LNSD, line 1: EIP-712 domain separator is an essential part of the signing schema. I miss it here.

Check this implementation for an example: https://github.com/edgeandnode/toolshed/blob/e7ea9a8209b7a960b4c3bd4e9a0dd2f2ea9019a2/thegraph-core/src/attestation.rs#L70-L79

service/src/database/dips.rs
```diff
@@ -0,0 +1,45 @@
...
+
+use std::collections::HashMap;
+
+use axum::async_trait;
+
+use crate::routes::dips::Agreement;
+
+#[async_trait]
+pub trait AgreementStore: Sync + Send {
+    async fn get_by_signature(&self, signature: String) -> anyhow::Result<Option<Agreement>>;
```
LNSD, line 12: I am considering using a UUIDv7 as the agreement identifier instead of the signature. It has some good properties that we can use to manage and trace data across the different DIPs' parties (i.e., DIPs Gateway/Dipper, Indexer, and future smart contracts).

Of course, these IDs will be only relevant on the indexer-facing side of the DIPs domain.

WDYT?

LNSD (commented):
Please, check my comments 🙂

service/src/database/dips.rs
LNSD: For example, as UUIDv7 is lexicographically sortable, we can quickly sort the agreements from oldest to newest without requiring a timestamp field.

dips/src/lib.rs
```diff
@@ -0,0 +1,388 @@
...
+        // should coincide with signer.
+        address payer;
+        // should coincide with indexer.
+        address payee;
+        // data service that will initiate payment collection.
+        address service;
+        // signer of the cancellation, can be signed by either party.
+        address cancellled_by;
+        // should only be usable within a limited period of time.
+        uint64 timestamp;
```
LNSD, line 73: I have followed the thread, and it is used in the cancellation request validation. What's the goal of this timestamp?

dips/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+alloy-sol-types = "=0.8.10"
+# alloy-signer = { version = "0.4", features = ["eip712"] }
+alloy-rlp = "0.3.9" 
+thegraph-core.workspace = true
+
+# alloy-signer = "0.4.2"
+# alloy-rlp = "0.3.8"
+
+# [dev-dependencies]
+# alloy-signer-local = "=0.4.0"
```
LNSD, line 19: Is there any reason to keep these lines commented?

service/src/database/dips.rs
```diff
@@ -0,0 +1,130 @@
...
+    ) -> anyhow::Result<Uuid> {
+        let bs = signed_cancellation.encode_vec();
+
+        sqlx::query!(
+            "UPDATE dips_agreements SET updated_at=now(), cancelled_at=now(), signed_cancellation_payload=$1 WHERE id=$2",
+bs,
+            id,
+        )
+        .execute(&self.pool)
+        .await?;
```
LNSD, line 126: I would return an error when there's no agreement in the DB with the provided ID. If I am correct, the query will not update anything and will return successfully.

service/src/routes/dips.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+
+    pub async fn cancel_agreement<'a>(
+        &self,
+        ctx: &'a Context<'_>,
+        id: String,
+        // data should be the signed voucher, eip712 signed, rlp and base64 encoded.
+        signed_request: String,
+    ) -> FieldResult<String> {
+        let store: &Arc<dyn AgreementStore> = ctx.data()?;
```
LNSD, line 176: Out of curiosity. Why did you choose dynamic dispatch over a generic argument?

service/src/service.rs
```diff
@@ -163,6 +178,32 @@ pub async fn run() -> anyhow::Result<()> {
         graph_node_query_base_url: &config.graph_node.query_url,
     });
 
+    let agreement_store: Arc<dyn AgreementStore> = Arc::new(InMemoryAgreementStore {
+        data: RwLock::new(HashMap::new()),
+    });
```
LNSD, line 183: If `InMemoryAgreementStore` implemented `Default::default()`, it would simplify the instantiation:

```suggestion
    let agreement_store: Arc<dyn AgreementStore> = Arc::new(InMemoryAgreementStore::default());
```

service/src/database/dips.rs
```diff
@@ -0,0 +1,130 @@
...
+        let item = match item {
+            Ok(item) => item,
+            Err(sqlx::Error::RowNotFound) => return Ok(None),
+            Err(err) => bail!(err),
+        };
+
+        let signed = SignedIndexingAgreementVoucher::decode(&mut item.signed_payload.as_ref())?;
+
+        Ok(Some(signed))
+    }
```
LNSD, line 86: I think this query is incorrect:
```suggestion
    async fn get_by_id(&self, id: Uuid) -> anyhow::Result<Option<SignedIndexingAgreementVoucher>> {
        let item = sqlx::query!(
            "SELECT * FROM dips_agreements WHERE id=$1",
            id,
        )
        .fetch_one(&self.pool)
        .await;

        let item = match item {
            Ok(item) => item,
            Err(sqlx::Error::RowNotFound) => return Ok(None),
            Err(err) => bail!(err),
        };

        let signed = SignedIndexingAgreementVoucher::decode(&mut item.signed_payload.as_ref())?;

        Ok(Some(signed))
    }
```

service/src/database/dips.rs
```diff
@@ -0,0 +1,130 @@
...
+            id,
+            agreement.signature.as_ref(),
+            bs,
+            protocol,
+            agreement.voucher.service.as_slice(),
+            agreement.voucher.payee.as_slice(),
+            agreement.voucher.payer.as_slice(),
+            now,
+            now,
+        )
```
LNSD, line 107: You can use the `timezone('UTC', now())` Postgres function to initialize the timestamps:
```suggestion
        sqlx::query!(
            "INSERT INTO dips_agreements VALUES ($1,$2,$3,$4,$5,$6,$7,timezone('UTC', now()),timezone('UTC', now()),null,null)",
            id,
            agreement.signature.as_ref(),
            bs,
            protocol,
            agreement.voucher.service.as_slice(),
            agreement.voucher.payee.as_slice(),
            agreement.voucher.payer.as_slice(),
        )
```

service/src/database/dips.rs
```diff
@@ -0,0 +1,130 @@
...
+    }
+    async fn cancel_agreement(
+        &self,
+        id: Uuid,
+        signed_cancellation: SignedCancellationRequest,
+    ) -> anyhow::Result<Uuid> {
+        let bs = signed_cancellation.encode_vec();
+
+        sqlx::query!(
+            "UPDATE dips_agreements SET updated_at=now(), cancelled_at=now(), signed_cancellation_payload=$1 WHERE id=$2",
```
LNSD, line 121: I am not sure if `now()` respects the _UTC_ timezone, I have been using the `timezone('UTC', now())` function:
```suggestion
            "UPDATE dips_agreements SET updated_at=timezone('UTC', now()), cancelled_at=timezone('UTC', now()), signed_cancellation_payload=$1 WHERE id=$2",
```

LNSD (dismissed):
In general, LGTM ✅

dips/src/lib.rs
mangas: essentially to prevent re-use or any variation of that.

service/src/database/dips.rs
mangas: Isn't that the correct outcome anyway? If the query doesn't update anything it's a noop. I don't have a strong preference, if you do I'll change it otherwise I think I'll leave that in for the moment

service/src/routes/dips.rs
mangas: Normally I find it easier to work with dynamic dispatch, generic arguments tend to propagate through the code very easily, I don't think the perf penalty on something like this is worth the trade-off.

service/src/service.rs
mangas: it does, I forgot to update it

LNSD (approved):
LGTM ✅

PR #339, opened 2024-10-02
fix: Add option to give either postgres_url or values individually

### LINKED ISSUE #316: Database configuration alternative
Some indexers would like to use user, password, host, port and database as an alternative to URL. 

We should provide as an alternative.

carlosvdr:
fixes #316

### DIFF
--- a/common/src/indexer_service/http/config.rs
+++ b/common/src/indexer_service/http/config.rs
@@ -10,6 +10,7 @@ use thegraph_core::{Address, DeploymentId};
 pub struct DatabaseConfig {
     pub postgres_url: String,
 }
+
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct SubgraphConfig {
     #[serde(default)]
--- a/common/src/indexer_service/http/config.rs
+++ b/common/src/indexer_service/http/config.rs
@@ -5,10 +5,41 @@ use std::net::SocketAddr;
 
 use serde::{Deserialize, Serialize};
 use thegraph_core::{Address, DeploymentId};
+use tracing::warn;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct DatabaseConfig {
     pub postgres_url: String,
+}
+impl DatabaseConfig {
+    pub fn format_db_config(
+        ps_url: Option<String>,
+        ps_host: Option<String>,
+        ps_pwd: Option<String>,
+        ps_port: Option<String>,
+        ps_user: Option<String>,
+        ps_db: Option<String>,
+    ) -> Self {
+        let db_config = (ps_url, ps_host, ps_pwd, ps_port, ps_user, ps_db);
+        match db_config {
+            (Some(url), ..) if !url.is_empty() => DatabaseConfig { postgres_url: url },
+            (None, Some(host), Some(pwd), Some(port), Some(user), Some(dbname)) => {
+                let postgres_url =
+                    format!("postgres://{}:{}@{}:{}/{}", user, pwd, host, port, dbname);
+                DatabaseConfig { postgres_url }
+            }
+            _ => {
+                warn!(
+                    "Some configuration values are missing for database values, please make sure you either \
+                    pass `postgres_url` or pass all the variables to connect to the database
+                ");
+                // This will eventually fail to connect
+                DatabaseConfig {
+                    postgres_url: String::new(),
+                }
+            }
+        }
+    }
 }
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
--- a/common/src/indexer_service/http/config.rs
+++ b/common/src/indexer_service/http/config.rs
@@ -5,43 +5,11 @@ use std::net::SocketAddr;
 
 use serde::{Deserialize, Serialize};
 use thegraph_core::{Address, DeploymentId};
-use tracing::warn;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct DatabaseConfig {
     pub postgres_url: String,
 }
-impl DatabaseConfig {
-    pub fn format_db_config(
-        ps_url: Option<String>,
-        ps_host: Option<String>,
-        ps_pwd: Option<String>,
-        ps_port: Option<String>,
-        ps_user: Option<String>,
-        ps_db: Option<String>,
-    ) -> Self {
-        let db_config = (ps_url, ps_host, ps_pwd, ps_port, ps_user, ps_db);
-        match db_config {
-            (Some(url), ..) if !url.is_empty() => DatabaseConfig { postgres_url: url },
-            (None, Some(host), Some(pwd), Some(port), Some(user), Some(dbname)) => {
-                let postgres_url =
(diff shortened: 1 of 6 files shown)


config/src/config.rs
```diff
@@ -149,6 +149,11 @@ pub struct IndexerConfig {
 #[cfg_attr(test, derive(PartialEq))]
 pub struct DatabaseConfig {
     pub postgres_url: Url,
+    pub postgres_host: String,
+    pub postgres_port: i32,
+    pub postgres_user: String,
+    pub postgres_password: String,
+    pub postgress_db: String,
```
gusinacio, line 156: You should turn this into an Enum with two variants: `Url { postgres_url: Url }` and `Splitten { .. }`. Also, password and port should be optional.

common/src/indexer_service/http/config.rs
```diff
@@ -5,11 +5,42 @@ use std::net::SocketAddr;
...
 use serde::{Deserialize, Serialize};
 use thegraph_core::{Address, DeploymentId};
+use tracing::warn;
 
 #[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct DatabaseConfig {
     pub postgres_url: String,
 }
+impl DatabaseConfig {
+    pub fn format_db_config(
```
gusinacio, line 15: Instead of a function here, `config` should have a `get_formatted_db_url(&self)-> Url`

gusinacio (changes requested):
It's also missing tap-gent configuration.

carlosvdr:
Will fix commit from fix - > to feat but dont want to force push until you have finished reviewing

config/minimal-config-example.toml
```diff
@@ -19,6 +19,13 @@ operator_mnemonic = "celery smart tip orange scare van steel radio dragon joy al
 # that is used by the `indexer-agent`. It is expected that `indexer-agent` will create
 # the necessary tables.
 postgres_url = "postgres://postgres@postgres:5432/postgres"
+# You can also use the following fields to specify the connection details separately.
+# either use `postgres_url` or the following fields.
+postgres_host = "postgres-host"
+postgres_port = 5432
+postgres_user = "user"
+postgres_password = "pwd"
+postgress_db = "postgres"
```
gusinacio, line 28: These configs are invalid, you should set just `host`, `port`, `user`, etc.

config/maximal-config-example.toml
```diff
@@ -33,6 +33,13 @@ port = 7300
 # that is used by the `indexer-agent`. It is expected that `indexer-agent` will create
 # the necessary tables.
 postgres_url = "postgres://postgres@postgres:5432/postgres"
+# You can also use the following fields to specify the connection details separately.
+# either use `postgres_url` or the following fields.
+postgres_host = "postgres-host"
+postgres_port = 5432
+postgres_user = "user"
+postgres_password = "pwd"
+postgress_db = "postgres"
```
gusinacio, line 42: same here

common/src/indexer_service/http/config.rs
```diff
@@ -10,7 +10,6 @@ use thegraph_core::{Address, DeploymentId};
 pub struct DatabaseConfig {
     pub postgres_url: String,
 }
-
```
gusinacio, line 13: we don't need this line diff

config/maximal-config-example.toml
gusinacio: you could use one style in the minimal config, and another in the maximal config so that means we are testing both. Don't forget to leave the comment "as alternative you can use this other one"

config/src/config.rs
```diff
@@ -147,8 +147,40 @@ pub struct IndexerConfig {
...
+        user: String,
+        password: Option<String>,
+        database: String,
+    },
+}
+impl DatabaseConfig {
+    pub fn get_formated_postgres_url(&self) -> Url {
+        match self {
+            DatabaseConfig::PostgresUrl { postgres_url } => postgres_url.clone(),
+            DatabaseConfig::PostgresVars {
```
gusinacio, line 167: This is a little nit-picking but you can use just `self` and not clone. This is useful if you don't plan to use DatabaseConfig twice.

config/src/config.rs
```diff
@@ -147,8 +147,40 @@ pub struct IndexerConfig {
...
+                host,
+                port,
+                user,
+                password,
+                database,
+            } => {
+                let postgres_url_str = format!("postgres://{}@{}/{}", user, host, database);
+                let mut postgres_url = Url::parse(&postgres_url_str).unwrap();
```
gusinacio, line 175: Could you use expect and add a nice message here like "Failed to parse database_url".

config/src/config.rs
```diff
@@ -147,8 +147,40 @@ pub struct IndexerConfig {
...
+                postgres_url
+                    .set_password(password.as_deref())
+                    .expect("url is wrong");
+                postgres_url.set_port(*port).expect("url is wrong");
```
gusinacio, line 179: Could you add more information like: "Failed to set password" and "Failed to set port".?

gusinacio (changes requested):
I loved this solution. Really elegant and uses the Rusty way. Thank you for this great code.

Here are some small nit-picks, typos and errors that I found on the documentation.

config/maximal-config-example.toml
carlosvdr: I prefer both having same style to make the whole thing cohesive

config/src/config.rs
carlosvdr: woops yes was going to add this and forgot

config/maximal-config-example.toml
carlosvdr: "as alternative you can use this other one", what did yopu meant by this?
There is already a comment 
```
# You can also use the following fields to specify the connection details separately.
# either use `postgres_url` or the following fields.
```

config/maximal-config-example.toml
```diff
@@ -33,6 +33,13 @@ port = 7300
 # that is used by the `indexer-agent`. It is expected that `indexer-agent` will create
 # the necessary tables.
 postgres_url = "postgres://postgres@postgres:5432/postgres"
```
gusinacio, line 35: Hey, for some reason our software doesn't have an error by having both. Could you check here https://serde.rs/container-attrs.html#deny_unknown_fields?

config/minimal-config-example.toml
```diff
@@ -19,6 +19,13 @@ operator_mnemonic = "celery smart tip orange scare van steel radio dragon joy al
 # that is used by the `indexer-agent`. It is expected that `indexer-agent` will create
 # the necessary tables.
 postgres_url = "postgres://postgres@postgres:5432/postgres"
+# You can also use the following fields to specify the connection details separately.
```
gusinacio, line 22: I understand you wanting to keep the same on both. Could you create a test if we have a `toml string` with the splitten config it's generating the expected PostgresVars variant?

gusinacio (changes requested):

gusinacio (approved):
LGTM

PR #340, opened 2024-10-02
fix: add warn where trigger_value is below 0.1 grt

### LINKED ISSUE #323: Create warning for too low trigger value
Some indexers are setting the trigger value too low, we consider that it must be at least 0.1 GRT, ideally over 1 GRT. Anything lower can cause issues with the heaviest allocation not returning any receipts.

We should add a warning in case it's lower than 0.1 GRT.

carlosvdr:
fixes #323

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -94,6 +94,28 @@ impl Config {
             ),
             _ => {}
         }
+        let grt_wei = self.tap.max_amount_willing_to_lose_grt.get_value();
+        let decimal = BigDecimal::from_u128(grt_wei).unwrap();
+        let divisor = &self.tap.rav_request.trigger_value_divisor;
+        let trigger_value = (decimal / divisor)
+            .to_u128()
+            .expect("Could not represent the trigger value in u128");
+        let minimum_recommended_for_max_willing_to_lose_grt = 0.1;
+        if trigger_value
+            < minimum_recommended_for_max_willing_to_lose_grt
+                .to_u128()
+                .unwrap()
+        {
+            warn!(
+                "Trigger value is too low,currently below 0.1 GRT. \
+                Please modify `max_amount_willing_to_lose_grt` or `trigger_value_divisor`.
+                To have a higher trigger value, ideally above 1 GRT \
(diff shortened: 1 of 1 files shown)


config/src/config.rs
```diff
@@ -94,6 +94,28 @@ impl Config {
...
+            < minimum_recommended_for_max_willing_to_lose_grt
+                .to_u128()
+                .unwrap()
+        {
+            warn!(
+                "Trigger value is too low,currently below 0.1 GRT. \
+                Please modify `max_amount_willing_to_lose_grt` or `trigger_value_divisor`.
+                To have a higher trigger value, ideally above 1 GRT \
+                Anything lower can cause issues with the heaviest \
+                allocation not returning any receipts.
```
gusinacio, line 114: The message could be: the system may constantly deny the sender.

config/src/config.rs
```diff
@@ -94,6 +94,28 @@ impl Config {
...
+            .to_u128()
+            .expect("Could not represent the trigger value in u128");
+        let minimum_recommended_for_max_willing_to_lose_grt = 0.1;
+        if trigger_value
+            < minimum_recommended_for_max_willing_to_lose_grt
+                .to_u128()
+                .unwrap()
+        {
+            warn!(
+                "Trigger value is too low,currently below 0.1 GRT. \
```
gusinacio, line 110: typo, space between comma and next letter

config/src/config.rs
```diff
@@ -94,6 +94,28 @@ impl Config {
...
+                Please modify `max_amount_willing_to_lose_grt` or `trigger_value_divisor`.
+                To have a higher trigger value, ideally above 1 GRT \
```
gusinacio, line 112: could you review the punctuation and capitalization of letters?

gusinacio (changes requested):

gusinacio:
Also, technically this is a feature. We are not fixing something that is broken, we are adding an alert functionality.

carlosvdr:
Hmm i dont like it being a feature but its closer to that rather than a fix yes

gusinacio (approved):
LGTM

PR #344, opened 2024-10-04
Ablity to use own enviornment variables in config files

### LINKED ISSUE #336: Add `envsubst` for config files
**Problem statement**
Indexers would like to use their own environment variable names within the config instead of relying on config paths, with separators.

**Expectation proposal**
If a config file contains `${VARIABLE}`, it should be replaced correctly with environment variables.

**Additional context**
This would help a lot with indexer migration from indexer-ts to indexer-rs.

taslimmuhammed:
Fixes #336 
I have tested the function alone in a separate environment, please do test it with entire project.


gusinacio:
Hello, just a few things before I review this PR.
1. Could you do a git rebase from our main branch?
2. Could you add some unit tests checking that this regex work?

Also, is there any reason to come up with your own regex over using a crate like [envsubst](https://docs.rs/envsubst/latest/envsubst/)?

Please don't change your solution, I'm just curious to know your thoughts.

gusinacio:
Down the `config.rs` file, there's a `mod test`. You can add a new `#[test]` function there and test it. You can use the other as examples.

taslimmuhammed:
thanks✌️, I'll do that

taslimmuhammed:
I have added test for only substitute_env_vars() alone not combining test_override_with_env, if required I'll add.

config/src/config.rs
```diff
@@ -71,17 +73,36 @@ impl Config {
...
             .extract()
             .map_err(|e| e.to_string())?;
-        config.0.validate()?;
 
+        config.0.validate()?;
         Ok(config.0)
     }
 
+    fn substitute_env_vars(content: String) -> Result<String, String> {
+        let re = Regex::new(r"\$\{([A-Z_][A-Z0-9_]*)\}").map_err(|e| e.to_string())?;
```
gusinacio, line 93: You can use an "expect" here instead of ? since the regex is constant.

config/src/config.rs
```diff
@@ -71,17 +73,36 @@ impl Config {
...
             .merge(Env::prefixed(prefix.get_prefix()).split("__"))
             .extract()
             .map_err(|e| e.to_string())?;
-        config.0.validate()?;
 
+        config.0.validate()?;
         Ok(config.0)
     }
 
+    fn substitute_env_vars(content: String) -> Result<String, String> {
```
gusinacio, line 92: This can be just "String" instead of Result

gusinacio (changes requested):
This looks good to me, just a small adjustment and this is ready to be merged.

config/src/config.rs
```diff
@@ -398,4 +418,45 @@ mod tests {
...
+    fn test_substitute_env_vars() {
+        // Set up environment variables
+        env::set_var("TEST_VAR1", "hello");
+        env::set_var("TEST_VAR2", "world");
+        // TEST_VAR3 not setup and should not be replaced by the function
+
+        let input = r#"
+            [section1]
+            key1 = "${TEST_VAR1}"
+            key2 = "${TEST_VAR2}"
```
gusinacio, line 434: Could you test some other keys and edge cases?

Like just "{TEST_VAR}" (without the $) or "${{TEST_VAR}}".

Some other cases like "${TEST_VAR:-default}" and "${TEST_VAR-default}". 

Just to have a clear understanding how the replacement works (you don't need to implement default values).

taslimmuhammed: got it[✌️]

config/src/config.rs
taslimmuhammed: Sure, sorry for that

config/src/config.rs
gusinacio: Hey, could you also set TEST_VAR to a value or update it to TEST_VAR1? Sorry I wasn't clear on my previous comment.

taslimmuhammed: still a little confused[🥲], did you mean ${TEST_VAR:-default} to a default value, or assign a value to it, as of it represents an variable which is not available in .env and wont be replaced

taslimmuhammed: if the corresponding env variables are not set, it should send an error warning to user right? rather than replacing it with default values

gusinacio: sorry for the confusion hahaha. I told in the comment to use TEST_VAR, but I want to understand what happens when there are default values and the env var is set. So I should have asked for TEAT_VAR1 (which is already defined)

taslimmuhammed: cool, I'll setup a modify test, so replacement will be checked

taslimmuhammed:
made more changes:-
- if the env variable is not found then it'll send error
- added a new test combining parsing and substitution of custom env vars, working good
please have a look, and BTW if you think the code is getting undauntedly lengthy lets use the package it's fine

taslimmuhammed:
also please note:-  in some places i have replaced std::env to just env

config/src/config.rs
```diff
@@ -71,17 +73,53 @@ impl Config {
...
         Ok(config.0)
     }
 
+    fn substitute_env_vars(content: String) -> Result<String, String> {
+        let reg = Regex::new(r"\$\{([A-Z_][A-Z0-9_]*)\}").map_err(|e| e.to_string())?;
+        let mut missing_vars = Vec::new();
+        let mut result = String::new();
+
+        for line in content.lines() {
+            if !line.trim_start().starts_with('#') {
```
gusinacio, line 97: Interesting that you are ignoring comment lines.

gusinacio (approved):
LGTM

taslimmuhammed:
Thanks alot 🙌

config/src/config.rs
taslimmuhammed: fortunately there was a variable in the comments of minimal_config, which failed the new test[✌️]

PR #346, opened 2024-10-04
feat: Add a warning in case senders are undenied manually from db

### LINKED ISSUE #338: Deny sender again if indexers remove from denylist
**Describe the bug**
We want to be extra sure that people MUST NOT remove a sender from the denylist. In case they do, deny it again and send a new message saying that by doing so, they are risking to lose their query fees money.

They can always just shutdown tap-agent, but this should not be our problem.

carlosvdr:
This also creates a new function which does the check for inital denial of a sender
fixes #338

### DIFF
--- a/tap-agent/src/agent/sender_account.rs
+++ b/tap-agent/src/agent/sender_account.rs
@@ -533,18 +533,7 @@ impl Actor for SenderAccount {
             message = ?message,
             "New SenderAccount message"
         );
-        // If state is denied and receivied new message, sender was removed manually from DB
-        if state.denied {
-            tracing::warn!(
-                "
-                No new messages should have been receieved, sender has been denied before. \
-                You SHOULD NOT remove a denied sender manually from the database. \
-                If you do so you are exposing yourself to potentially LOOSING ALL of your query
-                fee MONEY.
-                "
-            );
-            state.add_to_denylist().await;
-        }
+
         match message {
             SenderAccountMessage::UpdateRav(rav) => {
                 state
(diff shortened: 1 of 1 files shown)


tap-agent/src/agent/sender_account.rs
```diff
@@ -574,6 +575,19 @@ impl Actor for SenderAccount {
 
                 match receipt_fees {
                     ReceiptFees::NewReceipt(value) => {
+                        // If state is denied and receivied new receipt, sender was removed manually from DB
```
gusinacio, line 578: typo, `received`

tap-agent/src/agent/sender_account.rs
```diff
@@ -574,6 +575,19 @@ impl Actor for SenderAccount {
...
+                        if state.denied {
+                            tracing::warn!(
+                                "
+                                No new receipts should have been received, sender has been denied before. \
+                                You SHOULD NOT remove a denied sender manually from the database. \
+                                If you do so you are exposing yourself to potentially LOOSING ALL of your query
```
gusinacio, line 584: typo: `LOSING`

tap-agent/src/agent/sender_account.rs
```diff
@@ -574,6 +575,19 @@ impl Actor for SenderAccount {
...
+                                fee MONEY.
```
gusinacio, line 585: I liked the messages, do you think some **** around the capitalized letters can make it more flashy?

gusinacio (changes requested):
Really good code. just some typos

tap-agent/src/agent/sender_account.rs
carlosvdr: Okay ya

gusinacio (approved):
LGTM, simple and solves our issue!

PR #352, opened 2024-10-07
fix: Accept config to be through file or env vars

### LINKED ISSUE #337: Remove requirement for config file
**Problem statement**
It's possible to provide all config fields via environment variables, so we shouldn't require people to provide a file if they have all required config values.

carlosvdr:
fixes #337

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -70,17 +70,19 @@ impl ConfigPrefix {
 }
 
 impl Config {
-    pub fn parse(prefix: ConfigPrefix, filename: &PathBuf) -> Result<Self, String> {
+    pub fn parse(prefix: ConfigPrefix, filename: Option<&PathBuf>) -> Result<Self, String> {
         let config_defaults = include_str!("../default_values.toml");
 
-        let mut config_content = std::fs::read_to_string(filename)
+        let mut figment_config = Figment::new().merge(Toml::string(config_defaults));
+
+        if let Some(path) = filename {
+            let mut config_content = std::fs::read_to_string(path)
             .map_err(|e| format!("Failed to read config file: {}", e))?;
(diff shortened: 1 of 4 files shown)


carlosvdr:
So I didnt added it, as in the validation it should fail so didnt saw a reason to add it.

But maybe also add to first validate the env vars, and if they are not all okay to error out a file name is alaso required?

config/src/config.rs
```diff
@@ -68,12 +68,16 @@ impl ConfigPrefix {
 }
 
 impl Config {
-    pub fn parse(prefix: ConfigPrefix, filename: &PathBuf) -> Result<Self, String> {
+    pub fn parse(prefix: ConfigPrefix, filename: &Option<PathBuf>) -> Result<Self, String> {
```
gusinacio, line 71: Almost always it's not a good idea to pass a reference to an Option. In case you want a reference, you want to pass an Option to a reference.

`&Option<T>` -> `Option<&T>`.

service/src/service.rs
```diff
@@ -133,7 +133,7 @@ pub async fn run() -> anyhow::Result<()> {
         MainConfig::parse(indexer_config::ConfigPrefix::Service, &cli.config).map_err(|e| {
             error!(
                 "Invalid configuration file `{}`: {}",
-                cli.config.display(),
+                cli.config.unwrap_or_default().display(),
                 e
             );
             anyhow!(e)
```
gusinacio, line 139: In case there's a missing config field while parsing the config file, I'd like to let the user know that it's possible to use the `--config`. Please also update `tap-agent` accordingly.

gusinacio (changes requested):

gusinacio (approved):
LGTM

gusinacio (approved):

PR #356, opened 2024-10-08
feat: Make it possible to have a shared env var for srvice and tap

### LINKED ISSUE #335: Friendly environment variables setup
**Problem statement**
Currently, each service has its own tag (`INDEXER_SERVICE` and `TAP_AGENT`), but at the same time, they are using shared variables. This makes a lot harder to work with especially in some setups like docker-compose where yaml anchors would be a perfect fit to share config.

**Expectation proposal**
Create a backward-compatible way to provide a single environment variable that serves both services.
The new tag should be something like `INDEXER_`, suggestions are welcome.

carlosvdr:
fixes #335

### DIFF
--- a/config/src/config.rs
+++ b/config/src/config.rs
@@ -22,6 +22,8 @@ use url::Url;
 
 use crate::NonZeroGRT;
 
+const SHARED_PREFIX: &str = "INDEXER_";
+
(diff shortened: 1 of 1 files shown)


gusinacio (approved):
LGTM. Lovely work!

PR #369, opened 2024-10-13
refactor: replace eventuals with Tokio watch for disputes

### LINKED ISSUE #361: Drop eventuals in favor of Tokio watch + timers for `disputes`
Subissue for #333

This is a larger refactor that will update multiple parts of the code.

I suggest you start with dispute manager which is the smallest Eventual in the code.
https://github.com/graphprotocol/indexer-rs/blob/main/common/src/attestations/dispute_manager.rs

After this, we can have more PRs for  `attestation_signers`, `allocation` and `escrow_account`.

_Originally posted by @gusinacio in https://github.com/graphprotocol/indexer-rs/issues/333#issuecomment-2402390838_

taslimmuhammed:
Fixes #361 
tried fixing only disputes[caused a chain reaction], had to fix all related too, 
- didn't rename the pipe to handle[Although they are now handles, serves purpose of a pipe]
-  added wait for many tests in "common" dir.
- removed error check `.map_err(|e| CheckError::Retryable(e.into()))?`  in tap-agent signature

if any of the above violates current code/is wrong do tell, I'll fix, also there are so many unnecessary comments, I'll remove after review


gusinacio:
Wow, I didn't know it would cause this chain reaction. It turned out to be bigger than I thought, sorry for that.

taslimmuhammed:
No, I'm sorry I messed up, you only asked for changing the dispute variable right?, misunderstood, changed all the eventual variables in the errored files, if for now, you want change only on disputes do tell, I'll push it seperately.

gusinacio:
>No, I'm sorry I messed up, you only asked for changing the dispute variable right?, misunderstood, changed all the eventual variables in the errored files, if for now, you want change only on disputes do tell, I'll push it seperately.

Do you know if it's possible to split into 4 different PRs (assuming all of the PRs are gonna build/test correctly)?

It's currently too big and I'm scared of miss anything. In case this is not possible just let me know and I'll try to review carefully this big PR.

taslimmuhammed:
Sure, I'll do that, but I can only do that tomorrow, came out for a small interview .

common/src/attestations/dispute_manager.rs
```diff
@@ -25,27 +24,44 @@ struct DisputeManager;
...
-        move |err: String| {
-            warn!("Failed to query dispute manager: {}", err);
-
-            // Sleep for a bit before we retry
-            sleep(interval.div_f32(2.0))
-        },
-    )
+) -> Receiver<Address> {
+    let (tx, rx) = watch::channel(Address::default());
+    tokio::spawn(async move {
```
gusinacio, line 29: Hey, I was thinking about having 
```rust
type ValueWatcher<T> = Receiver<Option<T>>;
```

This way you don't need to do weird checks with `Default::default()`.

common/src/attestations/dispute_manager.rs
```diff
@@ -25,27 +24,44 @@ struct DisputeManager;
...
+        let mut time_interval = time::interval(interval);
```
gusinacio, line 30: I'd suggest updating the [missed_tick_behaviour](https://docs.rs/tokio/latest/tokio/time/struct.Interval.html#method.set_missed_tick_behavior) so you don't call this twice.

common/src/attestations/signers.rs
```diff
@@ -1,32 +1,39 @@
...
 
 /// An always up-to-date list of attestation signers, one for each of the indexer's allocations.
 pub fn attestation_signers(
-    indexer_allocations: Eventual<HashMap<Address, Allocation>>,
+    mut indexer_allocations: Receiver<HashMap<Address, Allocation>>,
     indexer_mnemonic: String,
     chain_id: ChainId,
-    dispute_manager: Eventual<Address>,
-) -> Eventual<HashMap<Address, AttestationSigner>> {
+    mut dispute_manager: Receiver<Address>,
```
gusinacio, line 17: Now I understand why you had all this rewrite at once, you are using it here.

common/src/attestations/signers.rs
```diff
@@ -1,32 +1,39 @@
...
+    let (tx, rx) = watch::channel(HashMap::new());
+
+    tokio::spawn(async move {
+        loop {
+            tokio::select! {
+                Ok(_) = indexer_allocations.changed() => {},
+                Ok(_) = dispute_manager.changed() => {},
+                else => break,
+            }
+
```
gusinacio, line 32: One suggestion to keep the dispute-manager isolated from `Eventual`s is to use:
```rust
let subscriber = eventual.subscribe();

select! {
    value = subscriber.read() = {}
    ...
```

This way you can rebuild the `join` without needing two eventuals.

I might be wrong on this one, let me know how this goes.

gusinacio (commented):
Hey, this is just a mini-review about the current work. It's in the right direction!

common/src/attestations/signers.rs
taslimmuhammed: was outside yesterday, thanks for the review BTW, I'll push the update today itself ✌️

common/src/attestations/dispute_manager.rs
taslimmuhammed: nice

common/src/attestations/dispute_manager.rs
taslimmuhammed: yeah, that'd be good, otherwise error checks were little messy

gusinacio:
> hi, I'm little new to this, does this mean you have reviewed the most of code?, cause I was going to just limit the PR to disputes_manger as of now

Nope, this is just a verification check to understand test coverage of your PR. This means what's the percentage of your code is being tested by tests.

You don't need to worry about it.

common/src/attestations/signers.rs
taslimmuhammed: Sorry for the delay[🥲], subscriber didn't work as expected, so had to use pipe to listen to the changes, which in turn will update a new watch::channel for indexer_allocations, hope this will be good as eventually we will be migrating all to tokio  watch

taslimmuhammed: BTW great to know eventuals was build at edge&node itself 🔥

taslimmuhammed: eventuals::join would only work if both variables are updated right?, so if you'd like we could modify it to work, when either one is updated

gusinacio: yes, exactly

gusinacio: actually, it updates when either one updates as long as both are defined. It only joins when both are Some

taslimmuhammed: Ok, I might have made some mistakes in testing the function, I'll send the updated pr asap

taslimmuhammed:
Hi, please have a look, it was responding to both variable changes when I tested in a local program.

common/src/attestations/dispute_manager.rs
```diff
@@ -25,27 +24,44 @@ struct DisputeManager;
...
+                let response = network_subgraph
+                    .query::<DisputeManager, _>(dispute_manager::Variables {})
+                    .await
+                    .map_err(|e| e.to_string())?;
+
+                response.map_err(|e| e.to_string()).and_then(|data| {
+                    data.graph_network
+                        .map(|network| network.dispute_manager)
+                        .ok_or_else(|| "Network 1 not found in network subgraph".to_string())
+                })
```
gusinacio, line 45: One cool thing about not needing eventuals anymore is that you don't need to convert to string each error.

Could you remove all of these `map_err`? We can use `anyhow::Error`.

common/src/attestations/signers.rs
```diff
@@ -15,47 +19,92 @@ pub fn attestation_signers(
...
+    tokio::spawn(async move {
+        //listening to the allocation eventual and converting them to reciever
+        //using pipe for updation
+        //for temporary pupose only
+        let (allocations_tx, mut allocations_rx) =
+            watch::channel(indexer_allocations.value().await.unwrap());
+        let _p1 = indexer_allocations.pipe(move |allocatons| {
+            let _ = allocations_tx.send(allocatons);
+        });
```
gusinacio, line 40: wow, this is really interesting.

common/src/attestations/signers.rs
```diff
@@ -15,47 +19,92 @@ pub fn attestation_signers(
...
+                    chain_id,
+                    attestation_signers_map,
+                    allocations_rx.clone(),
+                    dispute_manager_rx.clone(),
+                    signers_writer).await;
+                },
+                else=>{
+                    //something is wrong
+                    break;
                 }
```
gusinacio, line 64: I think you can implement this else with `panic!("dispute_manager_rx or allocations_rx was dropped")`.

common/src/attestations/signers.rs
```diff
@@ -15,47 +19,92 @@ pub fn attestation_signers(
...
+    dispute_manager_rx: Receiver<Option<Address>>,
+    mut signers_writer: EventualWriter<HashMap<Address, AttestationSigner>>,
+) -> EventualWriter<HashMap<Address, AttestationSigner>> {
+    let mut signers = attestation_signers_map.lock().await;
+    let allocations = allocations_rx.borrow().clone();
+    let dispute_manager = *dispute_manager_rx.borrow();
+    if dispute_manager.is_none() {
+        return signers_writer;
+    }
+    let dispute_manager = dispute_manager.unwrap();
```
gusinacio, line 85:
```rust
let Some(dispute_manager) = *dispute_manager_rx.borrow() else {
    return signers_writer;
};
```

common/src/attestations/signers.rs
```diff
@@ -15,47 +19,92 @@ pub fn attestation_signers(
...
-            signers.clone()
+    signers_reader
+}
+async fn modify_sigers(
+    indexer_mnemonic: Arc<String>,
+    chain_id: ChainId,
+    attestation_signers_map: &'static Mutex<HashMap<Address, AttestationSigner>>,
+    allocations_rx: Receiver<HashMap<Address, Allocation>>,
+    dispute_manager_rx: Receiver<Option<Address>>,
+    mut signers_writer: EventualWriter<HashMap<Address, AttestationSigner>>,
```
gusinacio, line 77: Question, is it possible for this to be `signers_writer: &mut EventualWriter<HashMap<Address, AttestationSigner>>,`?

gusinacio (changes requested):
Hey, I loved your solution. Here are just some small changes and questions.

suchapalaver (changes requested):

common/src/attestations/signers.rs
taslimmuhammed: Thanks a lot for the detailed review, I'll update and push in the morning, am not at home as of now.

taslimmuhammed: of course, &mut would have smoothly solved the ownership issues, of signers_writer, failed to consider it first time,   I have updated the code using it.

gusinacio (approved):
LGTM. Could you fix the issues with CI? We had a mini refactor updating the imports for Address to use `alloy::primitives`, I guess your merge didn't handle it.

common/src/attestations/signers.rs
```diff
@@ -15,47 +19,92 @@ pub fn attestation_signers(
...
-                    let signer = AttestationSigner::new(
-                        &indexer_mnemonic,
-                        allocation,
+    let (mut signers_writer, signers_reader) =
+        Eventual::<HashMap<Address, AttestationSigner>>::new();
+
+    tokio::spawn(async move {
+        //listening to the allocation eventual and converting them to reciever
+        //using pipe for updation
+        //for temporary pupose only
```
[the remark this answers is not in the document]
suchapalaver: Can we correct this, please @taslimmuhammed ?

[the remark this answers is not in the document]
gusinacio: if I'm not mistaken, it's done.

suchapalaver (approved):

taslimmuhammed:
> LGTM. Could you fix the issues with CI? We had a mini refactor updating the imports for Address to use `alloy::primitives`, I guess your merge didn't handle it.

sure, I had to resolve last conflict on github so couldn't run commands, BTW:-
- `cargo clippy --all-targets --all-features # for clippy`
- `cargo fmt # for fmt`
- `cargo sqlx prepare --workspace -- --all-targets # for SQL offline mode`
even after running all of this some PR checks usually fails, am I doing it wrong, or missing any commands?

taslimmuhammed:
Thank you  for taking the time to review and approve my PR. I really appreciate your feedback and support in getting this merged!

gusinacio:
>even after running all of this some PR checks usually fails, am I doing it wrong, or missing any commands?

We have two tests that usually fail:
- Lint PR: we use [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/), if your commit messages are not like that, it'll fail. Usually, I squash and set the correct message for the PR. If you start using them, they'll start to pass.
- Build and upload Docker image: it's a problem on our end, we shouldn't try to push docker images in case it's an external contributor.

Other than those two checks, all of the others should pass correctly.

common/src/attestations/dispute_manager.rs
```diff
@@ -7,8 +7,6 @@ use std::time::Duration;
 use thegraph_core::Address;
 use tokio::sync::watch::{self, Receiver};
 use tokio::time::{self, sleep};
-use alloy::primitives::Address;
```
gusinacio, line 6: Could you use `use alloy::primitives::Address;` instead of `use thegraph_core::Address;`

gusinacio (changes requested):

common/src/attestations/dispute_manager.rs
taslimmuhammed: oh , sorry

gusinacio (approved):
LGTM, thank you so much for the clean and beautiful code.

taslimmuhammed:
Really humbled by your kind words, BTW could you assign `attestation_signers` too to me

PR #374, opened 2024-10-14
chore: add ansi feature flag for subscriber

gusinacio:
If you try to compile directly `service` after compiling other components, you need to do `cargo clean && cargo build`. By adding this flag, you don't have to clean and rebuild


suchapalaver:
What is ansi?

suchapalaver (approved):

PR #377, opened 2024-10-15
feat: Move backoff to to the tracker to remove errors

### LINKED ISSUE #318: Backoff in Tracker per Allocation
Currently, backoff is being thrown as an error and it's becoming hard to understand what is the real error causing the backoff.

We want to create the backoff inside the tracker so that in case there's an error, that allocation is not selected until the backoff period is over.

carlosvdr:
fixes #318

### DIFF
--- a/tap-agent/src/agent/sender_account.rs
+++ b/tap-agent/src/agent/sender_account.rs
@@ -217,8 +217,23 @@ impl State {
             );
         };
         // we call and wait for the response so we don't process anymore update
-        let Ok((fees, rav)) = call!(allocation, SenderAllocationMessage::TriggerRAVRequest) else {
+        let Ok(rav_result) = call!(allocation, SenderAllocationMessage::TriggerRAVRequest) else {
             anyhow::bail!("Error while sending and waiting message for actor {allocation_id}");
+        };
+        let (fees, rav) = match rav_result {
+            Ok(ok_value) => {
+                self.rav_tracker.ok_rav_request(allocation_id);
+                ok_value
+            }
+            Err(err) => {
+                self.rav_tracker.failed_rav_backoff(allocation_id);
+                anyhow::bail!(
+                    "Error while requesting RAV for sender {} and allocation {}: {}",
+                    self.sender,
+                    allocation_id,
+                    err
+                );
+            }
         };
 
         let rav_value = rav.map_or(0, |rav| rav.message.valueAggregate);
--- a/tap-agent/src/agent/sender_account.rs
+++ b/tap-agent/src/agent/sender_account.rs
@@ -1097,21 +1097,19 @@ pub mod tests {
         next_unaggregated_fees_value: Arc<Mutex<u128>>,
         receipts: Arc<Mutex<Vec<NewReceiptNotification>>>,
     }
-    type RavRequestResultTuple = (Arc<AtomicU32>, Arc<Mutex<u128>>);
     impl MockSenderAllocation {
-        pub fn new_with_triggered_rav_request() -> (Self, Result<RavRequestResultTuple, ()>) {
+        pub fn new_with_triggered_rav_request() -> (Self, Arc<AtomicU32>, Arc<Mutex<u128>>) {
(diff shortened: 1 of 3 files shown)


tap-agent/src/agent/sender_account.rs
```diff
@@ -1082,20 +1097,21 @@ pub mod tests {
...
-                    triggered_rav_request: triggered_rav_request.clone(),
+                    triggered_rav_request,
                     receipts: Arc::new(Mutex::new(Vec::new())),
                     next_rav_value: Arc::new(Mutex::new(0)),
-                    next_unaggregated_fees_value: unaggregated_fees.clone(),
+                    next_unaggregated_fees_value: unaggregated_fees,
                 },
-                triggered_rav_request,
-                unaggregated_fees,
+                Ok(mock_rav_request_response),
```
suchapalaver, line 1114: Why are we returning a `Result` here? I couldn't an error we're trying to handle 🤔

tap-agent/src/agent/sender_fee_tracker.rs
```diff
@@ -40,6 +40,21 @@ pub struct SenderFeeTracker {
...
     blocked_addresses: HashSet<Address>,
+    failed_ravs: HashMap<Address, FailedRavInfo>,
+}
+
+#[derive(Debug, Clone)]
+pub struct FailedRavInfo {
+    failed_ravs_count: u32,
+    failed_rav_backoff_time: Instant,
+}
+impl Default for FailedRavInfo {
```
suchapalaver, line 51:
```suggestion

impl Default for FailedRavInfo {
```

suchapalaver:
@carlosvdr The main thing looks good to me but I'm going to let @gusinacio take a look 👍

tap-agent/src/agent/sender_account.rs
carlosvdr: Did so to make it more like the [real response](https://github.com/graphprotocol/indexer-rs/blob/f6ebdd8e1b1441e927cc202d9510ff8edba652c7/tap-agent/src/agent/sender_account.rs#L220), where it also responds with a result

tap-agent/src/agent/sender_fee_tracker.rs
```diff
@@ -288,4 +322,35 @@ mod tests {
         assert_eq!(tracker.get_total_fee_outside_buffer(), 0);
         assert_eq!(tracker.get_total_fee(), 0);
     }
+
+    #[test]
+    fn test_filtered_backed_off_allocations() {
+        let allocation_id_0 = address!("abababababababababababababababababababab");
+        let allocation_id_1 = address!("bcbcbcbcbcbcbcbcbcbcbcbcbcbcbcbcbcbcbcbc");
+        const BUFFER_WINDOW: Duration = Duration::from_secs(61);
```
gusinacio, line 330: Hey, I don't like that this is 61 seconds. This is sleeping, and it means that we need to wait 61 seconds for tests to pass.

Maybe you could update this to be really small, so it's not that problematic.

Also, could you update the name to BACKOFF_SLEEP_DURATION so it's clear what it does.

tap-agent/src/agent/sender_account.rs
```diff
@@ -1196,8 +1212,9 @@ pub mod tests {
         ActorRef<SenderAllocationMessage>,
         JoinHandle<()>,
     ) {
-        let (mock_sender_allocation, triggered_rav_request, next_unaggregated_fees) =
+        let (mock_sender_allocation, rav_request_result) =
             MockSenderAllocation::new_with_triggered_rav_request();
+        let (triggered_rav_request, next_unaggregated_fees) = rav_request_result.unwrap();
```
gusinacio, line 1217: You are unwrapping the result from `new_with_triggered_rav_request`. Isn't removing the result better?

gusinacio (changes requested):

tap-agent/src/agent/sender_account.rs
carlosvdr: yeah removed it afterall

PR #379, opened 2024-10-16
perf: use latest_rav to recalculate fees

gusinacio:


carlosvdr (approved):
nice 👍 , like what you did, LGTM

PR #381, opened 2024-10-16
perf: trigger rav request concurrently

### LINKED ISSUE #380: Allow multiple RAV requests to run at the same time
We currently have a system that performs only one RAV at a time. But it's possible to trigger multiple RAVs for multiple allocations at the same time. We can improve this.

gusinacio:
Fixes #380


carlosvdr (approved):
Code looks overall good with me, id like to have a call about it if possible

gusinacio (commented):
It's missing a way to block the allocation from triggering another rav, while a rav is being executed.

carlosvdr (approved):

PR #384, opened 2024-10-18
feat: add counter trigger for rav request

carlosvdr:


tap-agent/src/agent/sender_fee_tracker.rs
```diff
@@ -92,8 +109,8 @@ impl SenderFeeTracker {
     pub fn update(&mut self, id: Address, fee: u128) {
         if fee > 0 {
             // insert or update, if update remove old fee from total
-            if let Some(old_fee) = self.id_to_fee.insert(id, fee) {
-                self.total_fee -= old_fee;
+            if let Some(old_fee) = self.id_to_fee.insert(id, FeeCounter { fee, count: 0 }) {
+                self.total_fee -= old_fee.fee;
```
gusinacio, line 113: It needs to have an update here for `pub fn update(&mut self, id: Address, fee: u128, counter: usize)`. SQL query and UnaggregatedReceipts need to be updated as well.

gusinacio (changes requested):

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -417,6 +421,7 @@ impl SenderAllocationState {
                 .unwrap_or(BigDecimal::from(0))
                 .to_string()
                 .parse::<u128>()?,
+            counter: res.count.unwrap_or(0).to_string().parse::<usize>()?,
```
gusinacio, line 424: why do you need to_string().parse::<usize>()?

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -398,6 +400,7 @@ impl SenderAllocationState {
...
```
gusinacio, line 403: same here, why not just convert to usize?

tap-agent/src/agent/sender_allocation.rs
carlosvdr: tried doing just res.count.unwrap_or(0 as usize) but it would still be u64, got any suggestion?

gusinacio (commented):
I'd like to have metrics regarding the amount of receipts per allocation. This can be done on a following PR.

gusinacio (approved):
LGTM

PR #387, opened 2024-10-19
refactor: use tokio::watch for attestation_signers

### LINKED ISSUE #363: Drop eventuals in favor of Tokio watch + timers for `attestation_signers`
Subissue of #333

taslimmuhammed:
Fixes #363


common/src/indexer_service/http/request_handler.rs
```diff
@@ -155,10 +155,10 @@ where
         .map_err(IndexerServiceError::ReceiptError)?;
 
     // Check if we have an attestation signer for the allocation the receipt was created for
-    let signers = state
-        .attestation_signers
-        .value_immediate()
-        .ok_or_else(|| IndexerServiceError::ServiceNotReady)?;
```
taslimmuhammed, line 161: Have removed the  error check here, since the default value is `HashMap::new()`, hope that's ok?

gusinacio: Maybe what we could do to fix this is: Run the function and generate the attestation signers before creating the watcher.

Than you create the watcher that updates, this way we have an "Eventual that is already initialized".

gusinacio: Yes, you'd need to refactor not to pass the `&mut Sender`, and just return the `HashMap`.

You could run once before creating and then running inside the tokio task loop.

taslimmuhammed: that'd be cool✌️, thanks

taslimmuhammed:
Sorry, I think I have messed up a little bit I'll open new PR, I usually used to test on different repo,

taslimmuhammed:
Sorry, I think I have messed up a little, I'll send the updates in a new PR

taslimmuhammed:
Please take a look and tell me if any changes are needed

common/src/attestations/signers.rs
```diff
@@ -38,48 +39,56 @@ pub fn attestation_signers(
...
         });
-
+        modify_sigers(
+            Arc::new(indexer_mnemonic.clone()),
+            chain_id,
+            attestation_signers_map,
+            allocations_rx.clone(),
+            dispute_manager_rx.clone(),
+        )
+        .await;
```
gusinacio, line 49: My suggestion is to move this to before line 31, and then you can use it as input for `watch::channel`, so you don't initialize with an empty hashmap

gusinacio (changes requested):

common/src/attestations/signers.rs
taslimmuhammed: ok

taslimmuhammed: But modify_signers is an async fn as of now, I'll try to find a solution

gusinacio: Is it possible to turn `attestation_signers` into async?

taslimmuhammed: sure, I'll try

gusinacio: I have email notifications 😆

taslimmuhammed: Copy that✌️,

taslimmuhammed:
changed to async, should have done at first, sorry, always little reluctant to touch on legacy formats

taslimmuhammed:
Kinda sleepy now, I'm missing details, there seems to be an issue while calling the function above, no worries, I'll fix completely and let you know in the morning

taslimmuhammed:
Hi, please have a look when you're free, I believe this fixes the issue, tell me if this is not what you meant

common/src/attestations/signers.rs
```diff
@@ -1,85 +1,93 @@
...
-        // Using pipe for updation.
-        // For temporary pupose only.
-        let (allocations_tx, mut allocations_rx) =
-            watch::channel(indexer_allocations.value().await.unwrap());
+        // Actively listening to indexer_allocations to update allocations channel
+        // Temporary fix until the indexer_allocations is migrated to tokio watch
         let _p1 = indexer_allocations.pipe(move |allocatons| {
-            let _ = allocations_tx.send(allocatons);
+            allocations_tx.send(allocatons).unwrap();
         });
```
gusinacio, line 48: I think this could be moved to outside tokio::spawn. Probably right after line 29, so they are close to each other.

common/src/indexer_service/http/request_handler.rs
```diff
@@ -155,10 +155,10 @@ where
...
+    // Removing check for empty value since channel was created with an empty HashMap
+    //IndexerServiceError::ServiceNotReady
```
gusinacio, line 159: You can remove this comment.

gusinacio (changes requested):
Everything is really great! Just a few tweaks and we are ready to merge.

taslimmuhammed:
Thanks for the review, I'll send changes asap

taslimmuhammed:
Please have a look, I have made the changes that you asked, it's working good, BTW tell me if any more changes are required.

taslimmuhammed:
every merge has always been incredibly exciting, all thanks to you sir🫡

PR #396, opened 2024-10-23
fix: limit the max concurrent rav requests

gusinacio:


tap-agent/src/agent/sender_account.rs
```diff
@@ -637,46 +647,50 @@ impl Actor for SenderAccount {
...
-                        );
 
-                        state.rav_request_for_allocation(allocation_id).await
-                    }
-                    (_, true) => {
-                        tracing::debug!(
-                            total_fee_outside_buffer,
-                            trigger_value = state.config.tap.rav_request_trigger_value,
-                            "Total fee greater than the trigger value. Triggering RAV request"
+                let has_limit_available = state.adaptive_limiter.has_limit();
```
carlosvdr, line 651: I do not like much the name `has_limit_available`, id rather have something like `has_available_slots_for_requests` or thread or sth like that to make it more clear ?
What do you think 
I know its not parallel but maybe it gives the idea better

carlosvdr (approved):
lgtm

PR #397, opened 2024-10-23
fix: add backoff if could not find heaviest allocation

carlosvdr:


tap-agent/src/agent/sender_account.rs
```diff
@@ -673,10 +677,13 @@ impl Actor for SenderAccount {
...
                 // In case we fail, we want our actor to keep running
                 if let Err(err) = rav_result {
+                    state.backoff_info.fail();
                     tracing::error!(
                         error = %err,
                         "There was an error while requesting a RAV."
                     );
+                } else {
+                    state.backoff_info.ok();
                 }
```
gusinacio, line 687: This is not the place where we want our backoff to happen.

In general, the backoff.fail() should be on line 214, inside the ok_or for heaviest_allocation

tap-agent/src/agent/sender_account.rs
```diff
@@ -651,17 +656,16 @@ impl Actor for SenderAccount {
                     counter_greater_receipt_limit,
                     total_fee_greater_trigger_value,
                 ) {
-                    (true, _) => {
+                    (true, _) if !state.backoff_info.in_backoff() => {
```
gusinacio, line 659: we don't need to check if it's calling directly an allocation, only if we'd call a rav for heaviest allocation

gusinacio (approved):
LGTM but there's a test failing, could you check it?

gusinacio (changes requested):
Actually, the logic is wrong, we still want to check for counter even if it doesn't have backoff

gusinacio (approved):
LGTM

PR #403, opened 2024-10-24
docs(graphql): add docs for escrow account query

suchapalaver:
Thought I'd add some docs while I was working with these GraphQL queries...

---
[TAP-261](https://linear.app/semiotic/issue/TAP-261/document-escrow-account-query)


gusinacio (approved):
LGTM

PR #410, opened 2024-10-25
refactor: use config crate for tap-agent

gusinacio:


common/src/subgraph_client/client.rs
```diff
@@ -88,12 +88,23 @@ impl DeploymentDetails {
...
+        Self::for_graph_node_url(
+            Url::parse(graph_node_status_url)?,
+            Url::parse(graph_node_base_url)?,
+            deployment,
+        )
+    }
+
+    pub fn for_graph_node_url(
+        graph_node_status_url: Url,
+        graph_node_base_url: Url,
```
suchapalaver, line 101: You don't want to pass those in as a struct so there's no mixing them up?

config/src/grt.rs
```diff
@@ -4,10 +4,18 @@
 use bigdecimal::{BigDecimal, ToPrimitive};
 use serde::{de::Error, Deserialize};
 
-#[derive(Debug, PartialEq)]
+#[derive(Debug, PartialEq, Default)]
 pub struct NonZeroGRT(u128);
 
 impl NonZeroGRT {
+    pub fn new(value: u128) -> Result<Self, String> {
+        if value == 0 {
```
suchapalaver, line 12: Have you ever used something like https://doc.rust-lang.org/stable/std/num/type.NonZeroU128.html for this?

suchapalaver (dismissed):

common/src/subgraph_client/client.rs
gusinacio: There's a second refactor going on, I want to change as little as possible in common.

config/src/grt.rs
gusinacio: TIL, but this is used only in tests.

suchapalaver (approved):

PR #411, opened 2024-10-25
fix: refresh database before closing allocation

gusinacio:
With the new buffer, refreshing the unaggregated fees does not get to the latest state.

For it to happen, we need to refresh the whole database.

This is not supposed to happen because we only mark it as last after it was closed for 1 hour, but in case something goes wrong, we shouldn't miss anything.

### DIFF
--- a/tap-agent/src/agent/sender_allocation.rs
+++ b/tap-agent/src/agent/sender_allocation.rs
@@ -196,7 +196,7 @@ impl Actor for SenderAllocation {
                 Err(err) => {
                     error!(
                         error = %err,
-                        "There was an error while calculating last unaggregated receipts. Retrying in 30 seconds...");
+                        "There was an error while calculating the last unaggregated receipts. Retrying in 30 seconds...");
                     tokio::time::sleep(Duration::from_secs(30)).await;
                 }
             }
--- a/tap-agent/src/agent/sender_allocation.rs
+++ b/tap-agent/src/agent/sender_allocation.rs
@@ -154,7 +154,7 @@ impl Actor for SenderAllocation {
         }
 
         // update unaggregated_fees
-        state.unaggregated_fees = state.initialize_unaggregated_receipts().await?;
+        state.unaggregated_fees = state.refresh_all_database().await?;
 
         sender_account_ref.cast(SenderAccountMessage::UpdateReceiptFees(
             allocation_id,
@@ -187,6 +187,20 @@ impl Actor for SenderAllocation {
             allocation_id = %state.allocation_id,
             "Closing SenderAllocation, triggering last rav",
         );
+        loop {
+            match state.refresh_all_database().await {
+                Ok(value) => {
+                    state.unaggregated_fees = value;
+                    break;
+                }
+                Err(err) => {
+                    error!(
+                        error = %err,
+                        "There was an error while calculating last unaggregated receipts. Retrying in 30 seconds...");
+                    tokio::time::sleep(Duration::from_secs(30)).await;
+                }
+            }
+        }
         // Request a RAV and mark the allocation as final.
         while state.unaggregated_fees.value > 0 {
             if let Err(err) = state.request_rav().await {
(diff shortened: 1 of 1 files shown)


tap-agent/src/agent/sender_allocation.rs
```diff
@@ -187,6 +187,20 @@ impl Actor for SenderAllocation {
...
+            match state.refresh_all_database().await {
+                Ok(value) => {
+                    state.unaggregated_fees = value;
+                    break;
+                }
+                Err(err) => {
+                    error!(
+                        error = %err,
+                        "There was an error while calculating last unaggregated receipts. Retrying in 30 seconds...");
+                    tokio::time::sleep(Duration::from_secs(30)).await;
```
anirudh2, line 200: If `state.refresh_all_database` errors, this will log the error. If `state.refresh_all_database` continues to error, won't we just get stuck in an infinite loop? There's no fallback functionality is there?

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -187,6 +187,20 @@ impl Actor for SenderAllocation {
...
+        loop {
+            match state.refresh_all_database().await {
+                Ok(value) => {
+                    state.unaggregated_fees = value;
+                    break;
+                }
+                Err(err) => {
+                    error!(
+                        error = %err,
+                        "There was an error while calculating last unaggregated receipts. Retrying in 30 seconds...");
```
anirudh2, line 199: Would there be more detail in the logs than this to help investigate? This doesn't tell us what the error was, just where it happened.

anirudh2: Also, grammar-wise, did you mean to say, "There was an error while calculating **_the_** last unaggregated receipts. Retrying in 30 seconds..."? I'm not sure how to interpret what this error is telling me.

tap-agent/src/agent/sender_allocation.rs
```diff
@@ -352,7 +366,7 @@ impl SenderAllocationState {
         })
     }
 
-    async fn initialize_unaggregated_receipts(&self) -> Result<UnaggregatedReceipts> {
+    async fn refresh_all_database(&self) -> Result<UnaggregatedReceipts> {
```
anirudh2, line 369: I'm not a massive fan of this name. Is this only used to calculate unaggregated fees? If so, could we call it `calculate_unaggregated_fees`?

anirudh2 (changes requested):
Took a bit longer to get to this today than I'd hoped, but hopefully this helps!

tap-agent/src/agent/sender_allocation.rs
gusinacio: No, we are closing the allocation. There's nothing else to do other than retry again. Same happens for last rav, we need to aggregate all pending fees so we can mark it as last.

We don't have any fallback other than human interaction.

tap-agent/src/agent/sender_allocation.rs
gusinacio: The details are printed in line 198.

tap-agent/src/agent/sender_allocation.rs
gusinacio: We have two functions that recalculate unaggregated fees.

One recalculates up to the `last_id` in our state, and the other recalculates the whole database (where last_id < max::u64).

gusinacio: I just didn't want to have a `calculate_unaggregated_fee_for_whole_database()`, do you have any suggestion that can differentiate them better?

tap-agent/src/agent/sender_allocation.rs
gusinacio: Grammar-wise, I don't have strong opnions. gonna update it.

tap-agent/src/agent/sender_allocation.rs
gusinacio: done

tap-agent/src/agent/sender_allocation.rs
gusinacio: done

anirudh2 (approved):
Thanks, @gusinacio!

PR #420, opened 2024-10-29
refactor: remove clones and use static config in service

### LINKED ISSUE #414: Remove clones from config and use 'static Config in Service
Hey, I think you can `Box::leak` the configuration, that way you don't need to clone things.

_Originally posted by @gusinacio in https://github.com/graphprotocol/indexer-rs/pull/409#discussion_r1817154855_

shiyasmohd:
Fixes #414


gusinacio (approved):
LGTM

PR #430, opened 2024-10-29
Revert "fix: disable value check"

gusinacio:
Reverts graphprotocol/indexer-rs#429


anirudh2 (approved):
Simple revert of something we've already reviewed!

PR #409, opened 2024-10-25
refactor: use config crate for service

shiyasmohd:
This PR closes #402 

Changes in this PR
- Updated imports in `indexer_service.rs` & `service/src/service.rs`
- Delete `config.rs` from `common/indexer-service` & `service`
- Added a fn to get as `SocketAddr` from `MetricsConfig`

PR #438, opened 2024-10-30
fix: proper build graph-node query url

gusinacio:
We had a refactor(#409) that updated the duplication of code in configurations. But while doing that, status_url was misconfigured. With the latest refactor (#428), we crash early if we cannot start some services.

Indexer service on version 1.2.0 is not capable of starting up if you provide local deployment (graph-node + deployment_id)


anirudh2 (approved):
Talked offline about this PR.

PR #439, opened 2024-10-30
feat: add sender fee tracker metric

gusinacio:


tap-agent/src/agent/sender_account.rs
```diff
@@ -52,6 +52,12 @@ lazy_static! {
...
     .unwrap();
+    static ref SENDER_FEE_TRACKER: GaugeVec = register_gauge_vec!(
+        "tap_sender_fee_tracker_grt_total",
+        "Sender fee tracker metric",
+        &["sender"]
+    )
+    .unwrap();
     static ref INVALID_RECEIPT_FEES: GaugeVec = register_gauge_vec!(
         "tap_invalid_receipt_fees_grt_total",
         "Failed receipt fees",
```
suchapalaver, line 63: Why is there no need to pass the something like `&["sender"]` here as above?

gusinacio: There's, take a look on line 64

PR #464, opened 2024-11-06
fix: use request from gateway instead of serde req

### LINKED ISSUE #463: Attestation expect the query to be formatted
We are deserializing and serializing the request before signing. Some queries will fail because they are not formatted and the this process removes some whitespaces.

gusinacio:
Fixes #463


Theodus (dismissed):
LGTM

Theodus (approved):

PR #467, opened 2024-11-06
feat: Add grace period

carlosvdr:


common/src/tap/checks/value_check.rs
```diff
@@ -48,13 +50,16 @@ pub struct MinimumValue {
     cost_model_map: CostModelMap,
     global_model: GlobalModel,
     watcher_cancel_token: tokio_util::sync::CancellationToken,
+    updated_at: GracePeriod,
+    grace_period: u64,
```
gusinacio, line 54: Can we change this to `Duration`?

gusinacio (changes requested):

gusinacio (approved):
LGTM

PR #474, opened 2024-11-07
Fix dips

mangas:


.helix/languages.toml
```diff
@@ -0,0 +1,2 @@
+[language-server.rust-analyzer.environment]
+DATABASE_URL = "postgresql://postgres:postgres@127.0.0.1:5432"
```
gusinacio, line 2: Could you add .helix to .gitignore?

service/src/service.rs
```diff
@@ -163,15 +177,47 @@ pub async fn run() -> anyhow::Result<()> {
...
+                cancel_voucher_time_tolerance: cancellation_time_tolerance
+                    .unwrap_or(Duration::from_secs(5)),
+            },
+            EmptySubscription,
+        )
+        .data(agreement_store)
+        .data(prices)
+        .finish();
+
+        router = router.route("/dips", post_service(GraphQL::new(schema)));
```
gusinacio, line 212: this is nice!

tap-agent/Cargo.toml
```diff
@@ -37,10 +37,10 @@ ruint = { version = "1.12.3", features = [
...
 futures-util = { version = "0.3.28", default-features = false }
 jsonrpsee = { version = "0.24.0", features = ["http-client", "tracing"] }
-tap_aggregator = { git = "https://github.com/semiotic-ai/timeline-aggregation-protocol", rev = "eb8447e" }
 ractor = { version = "0.13", features = [
   "async-trait",
 ], default-features = false }
+tap_aggregator.workspace = true
 futures = { version = "0.3.30", default-features = false }
 
 [dev-dependencies]
```
gusinacio, line 46: Could you add this change in a separate `chore` commit?

gusinacio (changes requested):
This is nice work! Just a few suggestions to organize the CHANGELOG and the project in general.

.helix/languages.toml
mangas: Is there a reason we can't have this just work out of the box for helix users?

tap-agent/Cargo.toml
mangas: Not really because it is an atomic change, this is require for all the crates to compile with the same versions

gusinacio: you could have one `chore: update dependencies`, that is atomic and then the `feat: add dips support`.

Bumping tap_aggregator is not a feature to enable dips in tap-agent

.helix/languages.toml
gusinacio: There are multiple editors and multiple people using it. I personally use Neovim and my colleges use VSCode and Emacs. If we added config files for all editors, we'd have tons of folders and files that are not related to the code.

We usually use `.env` to set DATABASE_URL for our development, which is ignored, since there's no standard way to run the test database.  I personally use port 5433, so this would be an issue for me even if I used helix.

gusinacio (approved):
Nice work, lgtm!

PR #487, opened 2024-11-13
refactor: move graphql queries to own crate

gusinacio:


anirudh2 (approved):
Simple move. Nothing seems weird here

PR #513, opened 2024-11-25
refactor: use router builder

gusinacio:


crates/service/src/service.rs
```diff
@@ -68,64 +55,119 @@ pub async fn run() -> anyhow::Result<()> {
...
+            ),
         )
-        .data(agreement_store)
-        .data(prices)
-        .finish();
+        .await,
+    ))
+}
+
+pub async fn shutdown_signal() {
```
anirudh2, line 151: Please document this or rename it. Without reading the code, I assumed this would be sending a shutdown signal, not setting up for one and awaiting it.

crates/service/tests/router_test.rs
```diff
@@ -0,0 +1,128 @@
...
+        .build();
+
+    let app = router.create_router().await.unwrap();
+
+    let timestamp = SystemTime::now()
+        .duration_since(SystemTime::UNIX_EPOCH)
+        .expect("Time went backwards")
+        .as_nanos();
+
+    let receipt = create_signed_receipt(allocation.id, 0, timestamp as u64, 100).await;
```
anirudh2, line 108: Please name the arguments. I have no idea what the inputs are without reading the function signature.

crates/service/tests/router_test.rs
```diff
@@ -0,0 +1,128 @@
...
+        .tcp_nodelay(true)
+        .build()
+        .expect("Failed to init HTTP client");
+
+    let allocation = INDEXER_ALLOCATIONS
+        .values()
+        .collect::<Vec<_>>()
+        .pop()
+        .unwrap()
+        .clone();
```
anirudh2, line 37: Is there a reason to collect instead of `let allocation = INDEXER_ALLOCATIONS.values().next().unwrap().clone();`?

crates/service/tests/router_test.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    let (_allocations_tx, allocations) = watch::channel(test_assets::INDEXER_ALLOCATIONS.clone());
+
+    let graph_node_url = Url::parse(&mock_server.uri()).unwrap();
+
+    let router = ServiceRouter::builder()
+        .database(database)
+        .domain_separator(TAP_EIP712_DOMAIN.clone())
+        .http_client(http_client)
+        .graph_node(leak(GraphNodeConfig {
```
anirudh2, line 72: I am kinda concerned by the number of `leak`s that I see in this PR. Could you talk me through why we need these? Could we refactor such that we don't need `'static` lifetimes?

anirudh2: I see you replaced some of them with `Arc` in your last commit, but there are a lot remaining.

anirudh2 (changes requested):

crates/service/tests/router_test.rs
gusinacio: All of these are from configuration files. Previously we had just `&'static Config` but this was really hard to test since you need to mock a bunch of config values that are not used.

I decided to use just the references needed for the Router. I'll try to refactor those to use just the reference instead of static references.

The reason why this is not an issue is because at runtime we already have config static.

There are 2 differences between a leaked reference vs Arcs that live the whole program:
- Arcs have the overhead of cloning
- leaked objects never run the Drop trait

The only reason I changed Manager from static to Arc is because we need to run drop and close the database connections, this got clearer with tests.

crates/service/src/service.rs
```diff
@@ -68,64 +55,119 @@ pub async fn run() -> anyhow::Result<()> {
...
+    )
+    .await;
+
+    // Establish Database connection necessary for serving indexer management
+    // requests with defined schema
+    // Note: Typically, you'd call `sqlx::migrate!();` here to sync the models
+    // which defaults to files in  "./migrations" to sync the database;
+    // however, this can cause conflicts with the migrations run by indexer
+    // agent. Hence we leave syncing and migrating entirely to the agent and
+    // assume the models are up to date in the service.
```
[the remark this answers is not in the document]
gusinacio: https://github.com/graphprotocol/indexer-rs#dependency-choices

crates/service/src/tap/checks/value_check.rs
```diff
@@ -514,16 +518,26 @@ mod tests {
             "Should require minimal value"
         );
 
-        let signed_receipt =
-            create_signed_receipt(ALLOCATION_ID, u64::MAX, u64::MAX, 500000000000000).await;
+        let signed_receipt = create_signed_receipt(
+            SignedReceiptRequest::builder()
+                .value(500000000000000)
```
anirudh2, line 523: You already assigned this magic number to a var `minimal_value`. Why not reuse that?

crates/monitor/src/attestation.rs
```diff
@@ -11,13 +11,15 @@ use thegraph_core::{Address, ChainId};
 use tokio::sync::watch::Receiver;
 use tracing::warn;
 
+pub type AttestationWatcher = Receiver<HashMap<Address, AttestationSigner>>;
+
 /// An always up-to-date list of attestation signers, one for each of the indexer's allocations.
 pub fn attestation_signers(
     indexer_allocations_rx: Receiver<HashMap<Address, Allocation>>,
```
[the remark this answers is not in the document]
anirudh2: You thumbs-up'ed this, but it still looks pending.

anirudh2 (changes requested):

anirudh2 (approved):
Thanks, @gusinacio!

PR #515, opened 2024-11-27
chore: bump rust version to 1.81

gusinacio:


anirudh2 (changes requested):
Clippy is failing.

anirudh2 (changes requested):
The clippy commit should be a `test`. I'd also be okay with `refactor`. `chore` is for something else!

anirudh2 (approved):
Thanks, Gustavo!

PR #516, opened 2024-11-27
test: remove sleeps from tests

gusinacio:
Signed-off-by: Gustavo Inacio <gustavo@semiotic.ai>


anirudh2 (approved):
This is great! I think using `actor.notify` is a pretty natural way to remove the sleeps. Nice one!

PR #522, opened 2024-12-01
feat: add graph-indexed header in subgraph query response

### LINKED ISSUE #494: Return syncing status (latest block synced) in header of subgraph query responses
**Problem statement**

Currently the Gateway uses query rewrites to keep track of block numbers on each chain.

**Expectation proposal**

Include header with `latestSyncedBlockNumber` and chainHeadBlock information: `chainHeadBlocknumber`, `chainheadBlockHash`, and `chainHeadBlocktimestamp`.

shiyasmohd:
Fixes #494

### DIFF
--- a/crates/service/src/routes/request_handler.rs
+++ b/crates/service/src/routes/request_handler.rs
@@ -42,6 +42,7 @@ pub async fn request_handler(
             value.to_str().map(|value| value == "true").unwrap_or(false)
         });
 
+    let graph_indexed = response.headers().get(GRAPH_INDEXED).cloned();
     let body = response
         .text()
         .await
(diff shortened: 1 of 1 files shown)


Theodus (approved):
I literally implemented this just now assuming no one did it yet, and it looks the same. So I'll give this an approval :+1:

crates/service/src/routes/request_handler.rs
```diff
@@ -54,5 +55,10 @@ pub async fn request_handler(
     let mut response = Response::new(body);
     response.extensions_mut().insert(attestation_input);
 
+    let graph_indexed = response.headers().get(GRAPH_INDEXED).cloned();
```
Theodus, line 58: This line needs to be moved above where `let mut response = Response::new(body);` aliases `response` away from the graph-node response.

Theodus (changes requested):

shiyasmohd:
@Theodus Thanks for the review. Pushed the changes

Theodus (approved):

PR #529, opened 2024-12-03
refactor: replace dips route with grpc server

mangas:


gusinacio (dismissed):
Looks like it won't affect current system. LGTM

gusinacio (approved):

PR #577, opened 2025-01-15
fix: retain headers in attestation middleware

Theodus:
The indexer-service has not been properly forwarding the `graph-indexed` header because the attestation middleware discarded all headers set by the handler.


gusinacio (approved):
LGTM

PR #578, opened 2025-01-16
feat: remove jsonrpc client in favor of grpc client

pedrohba1:


crates/tap-agent/src/agent/sender_account.rs
```diff
@@ -603,9 +604,14 @@ impl Actor for SenderAccount {
...
-        let sender_aggregator = HttpClientBuilder::default()
-            .request_timeout(config.rav_request_timeout)
-            .build(&sender_aggregator_endpoint)?;
+        let endpoint = Endpoint::new(sender_aggregator_endpoint.to_string())
+            .unwrap()
+            .connect_timeout(config.rav_request_timeout);
+
+        let sender_aggregator = TapAggregatorClient::connect(endpoint)
+            .await
+            .unwrap()
```
gusinacio, line 613: Could you properly handle the error instead of unwrapping it?

crates/tap-agent/src/agent/sender_allocation.rs
```diff
@@ -592,20 +589,25 @@ impl SenderAllocationState {
...
-                    )
+                    .aggregate_receipts(rav_request)
+                    // .request(
+                    //     "aggregate_receipts",
+                    //     rpc_params!(
+                    //         "0.0", // TODO: Set the version in a smarter place.
+                    //         valid_receipts,
+                    //         previous_rav
+                    //     ),
+                    // )
```
gusinacio, line 607: Can you remove these comment lines?

gusinacio (changes requested):

gusinacio:
One test is running forever, could you check if you are properly aborting the task?

.github/workflows/tests.yml
```diff
@@ -58,8 +58,10 @@ jobs:
         run: cargo sqlx migrate run
       - name: Check that the sqlx prepared query metadata is up-to-date
         run: cargo sqlx prepare --workspace --check -- --all-targets --all-features
-
-  clippy:
+      - name: Start mock server
+      run: |
+            cargo run --bin mock_tap_aggregator_server_runner &
```
pedrohba1, line 63: So, I've added this so there is less changes and boilerplate code when testing. Otherwise, a lot of tests would require running the mocked gRPC server and shutting it down in the test function. This is because unlike 
 HttpClientBuilder::default` that was being used before, `TapAggregatorClient` does not have a builder and just uses `connect()`. And this fails the tests because there was no server in the dummy endpoint being used.

crates/tap-agent/src/agent/sender_accounts_manager.rs
```diff
@@ -727,8 +730,8 @@ mod tests {
                 escrow_subgraph: get_subgraph_client().await,
                 network_subgraph: get_subgraph_client().await,
                 sender_aggregator_endpoints: HashMap::from([
-                    (SENDER.1, Url::parse("http://localhost:8000").unwrap()),
-                    (SENDER_2.1, Url::parse("http://localhost:8000").unwrap()),
+                    (SENDER.1, Url::parse(SENDER_DUMMY_ADDR).unwrap()),
```
pedrohba1, line 733: I've changed this to the `DUMMY_ADDR`. Tests don't fail, but please, check if this is the intended behaviour.

crates/tap-agent/src/bin/mock_tap_aggregator_server_runner.rs
pedrohba1, line 1: This is the server that is ran for the tests only.

pedrohba1:
The test that hangs is in `agent::sender_allocation::tests::test_close_allocation_with_pending_fees`, not sure why yet. But it does hang right in the  `sender_allocation.stop_and_wait(None, None).await.unwrap();`

crates/tap-agent/src/bin/mock_tap_aggregator_server_runner.rs
```diff
@@ -0,0 +1,74 @@
...
+pub struct MockTapAggregatorService;
+
+#[tonic::async_trait]
+impl TapAggregator for MockTapAggregatorService {
+    async fn aggregate_receipts(
+        &self,
+        _request: Request<RavRequest>,
+    ) -> Result<Response<RavResponse>, Status> {
+        // Do nothing, just return an unimplemented response
+        Err(Status::unimplemented("This is a mock service"))
```
gusinacio, line 20: This whole file for me doesn't make sense, you can just use the same thing that we are using for older tests: `tap_aggregator::server::run_server`

justfile
```diff
@@ -11,6 +11,9 @@ deps:
 url:
     @echo export DATABASE_URL=postgresql://postgres:postgres@127.0.0.1:5432
 
+test: 
+    cargo run --bin mock_tap_aggregator_server_runner & cargo test
+
```
gusinacio, line 16: This isn't needed

crates/tap-agent/src/agent/sender_allocation.rs
```diff
@@ -1256,10 +1258,6 @@ pub mod tests {
             message_receiver.recv().await.unwrap(),
             SenderAccountMessage::UpdateReceiptFees(_, ReceiptFees::RavRequestResponse(_))
         ));
-
```
gusinacio, line 1259: You need to do `handle.abort()` here, otherwise the test is gonna run forever.

.github/workflows/tests.yml
```diff
@@ -131,6 +131,8 @@ jobs:
         run: cargo install sqlx-cli --no-default-features --features postgres
       - name: Run the test sqlx migrations
         run: cargo sqlx migrate run
+      - name: Start mock server
+        run: cargo run --bin mock_tap_aggregator_server_runner &
```
gusinacio, line 135: Remove the this from the tests, it's not needed.

crates/tap-agent/src/agent/sender_account.rs
```diff
@@ -603,9 +604,17 @@ impl Actor for SenderAccount {
...
+        let endpoint = Endpoint::new(sender_aggregator_endpoint.to_string())
+            .expect("Failed to create an endpoint for the sender aggregator")
+            .connect_timeout(config.rav_request_timeout);
+
+        let sender_aggregator = TapAggregatorClient::connect(endpoint.clone())
+            .await
+            .expect(&format!(
+                "Failed to connect to the TapAggregator endpoint '{}'",
+                endpoint.uri()
+            ))
```
gusinacio, line 616: Using expecte is fair, but I'd rather just use `.with_context("Failed to connect to the TapAggregator endpoint")?` and let anyhow handle this up in the chain.

crates/tap-agent/src/agent/sender_accounts_manager.rs
gusinacio: Ok, so now we need the addr to up while starting a `SenderAccount`?

crates/tap-agent/src/agent/sender_account.rs
gusinacio: It's interesting that gRPC now keeps a connection open instead of knowing that the aggregator is down only while trying to request.

This makes tests a little different since we need a gRPC server up for them to connect. I recommend taking a look in how we create tests that need aggregator and replicate the same.

gusinacio (changes requested):

gusinacio:
> The test that hangs is in `agent::sender_allocation::tests::test_close_allocation_with_pending_fees`, not sure why yet. But it does hang right in the `sender_allocation.stop_and_wait(None, None).await.unwrap();`

When you close an allocation, it tries to create a last_rav by aggregating all receipts until they are all aggregated.

It looks like when you call `stop_and_wait` it's hanging trying to aggregate receipts, you need to provide a real aggregator for that test, probably.

pedrohba1:
> > The test that hangs is in `agent::sender_allocation::tests::test_close_allocation_with_pending_fees`, not sure why yet. But it does hang right in the `sender_allocation.stop_and_wait(None, None).await.unwrap();`
> 
> When you close an allocation, it tries to create a last_rav by aggregating all receipts until they are all aggregated.
> 
> It looks like when you call `stop_and_wait` it's hanging trying to aggregate receipts, you need to provide a real aggregator for that test, probably.
> 
> We were previously mocking the response for the aggregator server, but since we cannot mock HTTP responses with gRPC you need to fix the test to use our new aggregator.

I found this, which seems that we can use to mock the gRPC aggregator:

https://crates.io/crates/wiremock-grpc

crates/tap-agent/src/agent/sender_accounts_manager.rs
pedrohba1: I've changed it to use the addresses of the generated servers

gusinacio:
Closed in favor of #583

PR #593, opened 2025-01-30
refactor: add horizon support

gusinacio:
- Patch cargo with git and rev for unreleased crate
- Rename TapReceipt to TapHeader
- New service::tap::TapReceipt which represents both v1 and v2
- Update justfile

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -54,7 +54,7 @@ bigdecimal = "0.4.3"
 build-info = "0.0.39"
 tap_core = { version = "3.0.0", default-features = false }
 tap_aggregator = { version = "0.4.0", default-features = false }
-tap_graph = { version = "0.1.0", default-features = false }
+tap_graph = { version = "0.2.0", features = ["v2"] }
 tracing-subscriber = { version = "0.3", features = [
     "json",
     "env-filter",
@@ -81,3 +81,15 @@ prost = "0.13.4"
 prost-types = "0.13.3"
 dipper-rpc = { git = "https://github.com/edgeandnode/dipper/", rev = "c8700e2", default-features = false }
 tonic-build = "0.12.3"
+
+[patch.crates-io.tap_core]
+git = "https://github.com/semiotic-ai/timeline-aggregation-protocol"
+rev = "dbae001"
+
+[patch.crates-io.tap_aggregator]
+git = "https://github.com/semiotic-ai/timeline-aggregation-protocol"
+rev = "dbae001"
+
+[patch.crates-io.tap_graph]
+git = "https://github.com/semiotic-ai/timeline-aggregation-protocol"
+rev = "dbae001"
--- a/crates/service/src/middleware/auth/tap.rs
+++ b/crates/service/src/middleware/auth/tap.rs
@@ -20,18 +20,20 @@ use tap_core::{
     manager::{adapters::ReceiptStore, Manager},
     receipt::Context,
 };
-use tap_graph::{ReceiptAggregateVoucher, SignedReceipt};
+use tap_graph::ReceiptAggregateVoucher;
 use tower_http::auth::AsyncAuthorizeRequest;
 
-use crate::{error::IndexerServiceError, middleware::prometheus_metrics::MetricLabels};
+use crate::{
+    error::IndexerServiceError, middleware::prometheus_metrics::MetricLabels, tap::TapReceipt,
+};
 
 /// Middleware to verify and store TAP receipts
 ///
 /// It also optionally updates a failed receipt metric if Labels are provided
 ///
-/// Requires SignedReceipt, MetricLabels and Arc<Context> extensions
+/// Requires TapReceipt, MetricLabels and Arc<Context> extensions
 pub fn tap_receipt_authorize<T, B>(
-    tap_manager: Arc<Manager<T, SignedReceipt, ReceiptAggregateVoucher>>,
+    tap_manager: Arc<Manager<T, TapReceipt, ReceiptAggregateVoucher>>,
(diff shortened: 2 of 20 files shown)


crates/service/src/tap/receipt.rs
```diff
@@ -0,0 +1,74 @@
...
+    receipt::{WithUniqueId, WithValueAndTimestamp},
+    signed_message::SignatureBytes,
+};
+use thegraph_core::alloy::{dyn_abi::Eip712Domain, primitives::Address, signers::Signature};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TapReceipt {
+    V1(tap_graph::SignedReceipt),
+    V2(tap_graph::v2::SignedReceipt),
+}
```
suchapalaver, line 14: For my edification, why would this not be in, say `tap_core` or some common library. I'm not saying tap-core is the right place just wondering about the common libs for tap 🙏

suchapalaver, line 14: For my edification, why would this not be in, say `tap_core` or some common library? I'm not saying tap-core is the right place just wondering about the common libs for tap 🙏

crates/service/src/tap/receipt.rs
```diff
@@ -0,0 +1,74 @@
...
+impl WithUniqueId for TapReceipt {
+    type Output = SignatureBytes;
+
+    fn unique_id(&self) -> Self::Output {
+        match self {
+            TapReceipt::V1(receipt) => receipt.unique_id(),
+            TapReceipt::V2(receipt) => receipt.unique_id(),
+        }
+    }
+}
```
suchapalaver, line 74: looks really nice!

suchapalaver, line 74: looks really nice!

crates/service/src/tap/receipt_store.rs
```diff
@@ -121,8 +120,8 @@ pub struct DatabaseReceipt {
 impl DatabaseReceipt {
     fn from_receipt(receipt: CheckingReceipt, separator: &Eip712Domain) -> anyhow::Result<Self> {
         let receipt = receipt.signed_receipt();
-        let allocation_id = receipt.message.allocation_id.encode_hex();
-        let signature = receipt.signature.as_bytes().to_vec();
+        let allocation_id = receipt.allocation_id().encode_hex();
+        let signature = receipt.signature().as_bytes().to_vec();
```
suchapalaver, line 124: Nice to have these 👍

crates/service/src/tap/receipt.rs
gusinacio: This should probably go into `tap_graph` in the near future.

Cargo.toml
```diff
@@ -80,3 +80,17 @@ tonic = { version = "0.12.3", features = ["tls-roots", "gzip"] }
...
+rev = "dbae001"
+
+[patch.crates-io.tap_aggregator]
+git = "https://github.com/semiotic-ai/timeline-aggregation-protocol"
+rev = "dbae001"
+
+
+[patch.crates-io.tap_graph]
+git = "https://github.com/semiotic-ai/timeline-aggregation-protocol"
+rev = "dbae001"
```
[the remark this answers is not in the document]
gusinacio: This is pointing to main branch and should be removed as soon as we have the next release.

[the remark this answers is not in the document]
gusinacio: This is pointing to main branch and should be removed as soon as we have the next release, likely in the following PRs.

suchapalaver (approved):

crates/service/src/tap/receipt.rs
```diff
@@ -0,0 +1,74 @@
...
+    signed_message::SignatureBytes,
+};
+use thegraph_core::alloy::{dyn_abi::Eip712Domain, primitives::Address, signers::Signature};
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum TapReceipt {
+    V1(tap_graph::SignedReceipt),
+    V2(tap_graph::v2::SignedReceipt),
+}
+
```
carlosvdr, line 15: I dont see anything bad in here but to understand, why is it that every match is the same, is it expected that V1 and V2 will have different outcomes for their values? and this is just to have the "skeleton" for future updates?

gusinacio: They have the same fields, but v2 has a few more fields and thus need to be handled different for store and read operations.

crates/service/src/middleware/auth/tap.rs
```diff
@@ -41,11 +43,11 @@ pub fn tap_receipt_authorize<T, B>(
...
        + Send
 where
-    T: ReceiptStore<SignedReceipt> + Sync + Send + 'static,
+    T: ReceiptStore<TapReceipt> + Sync + Send + 'static,
     B: Send,
 {
-    move |request: Request<B>| {
-        let receipt = request.extensions().get::<SignedReceipt>().cloned();
+    move |mut request: Request<B>| {
+        let receipt = request.extensions_mut().remove::<TapReceipt>();
```
carlosvdr, line 50: Im not getting here why we are using remove instead of get, is it due to it now being mut?

carlosvdr (approved):

crates/service/src/middleware/auth/tap.rs
gusinacio: I was cloning here before, but I can simply remove and use the value since the receipt is not used anywhere else.

carlosvdr:
LGTM

suchapalaver (approved):

PR #594, opened 2025-01-30
refactor: add horizon support to tap-agent

gusinacio:


crates/indexer-receipt/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+            _ => None,
+        }
+    }
+
+    pub fn get_v2_receipt(&self) -> Option<&tap_graph::v2::SignedReceipt> {
+        match self {
+            TapReceipt::V2(receipt) => Some(receipt),
+            _ => None,
+        }
+    }
```
suchapalaver, line 116: Why is this necessary with the `v1` `v2` separation?

crates/tap-agent/src/test.rs
```diff
@@ -318,7 +329,10 @@ pub async fn store_batch_receipts(
     let mut values = Vec::with_capacity(receipts_len);
 
     for receipt in receipts {
-        let receipt = receipt.signed_receipt();
+        let receipt = match receipt.signed_receipt() {
+            TapReceipt::V1(receipt) => receipt,
+            TapReceipt::V2(_) => unimplemented!("V2 receipts not supported"),
```
suchapalaver, line 334: Is there a way we would want to guard against someone sending a V2 before we're ready? I was thinking of an alternative here where instead of panicking `continue` after logging a warning.

crates/tap-agent/src/test.rs
```diff
@@ -364,6 +378,16 @@ pub async fn store_batch_receipts(
 }
 
 pub async fn store_invalid_receipt(
+    pgpool: &PgPool,
+    signed_receipt: &TapReceipt,
+) -> anyhow::Result<u64> {
+    match signed_receipt {
+        TapReceipt::V1(signed_receipt) => store_invalid_receipt_v1(pgpool, signed_receipt).await,
+        TapReceipt::V2(_) => unimplemented!("V2 not supported"),
```
suchapalaver, line 386: Same thought here, why panic instead of returning an error?

gusinacio: It's easier to do Search All and look for unimplemented!() calls, also these are test functions.

No receipt should be V2 if we don't parse V2 receipts in the header.

crates/indexer-receipt/src/lib.rs
gusinacio: just for Aggregation, it's especially important that we can only aggregate V1 Receipts into V1 Ravs and for V2 as well.

suchapalaver (approved):

crates/indexer-receipt/src/lib.rs
suchapalaver: Doesn't the enum handle that? I'm trying to understand how a V2 receipt could match as V1.

crates/tap-agent/src/test.rs
carlosvdr: I guess at this point is okay to panic since no one is expected to send v2, perhaps when v2 is a possibility we def dont want to panic by then

crates/indexer-receipt/src/lib.rs
gusinacio: How do you suggest doing it?

carlosvdr (approved):

crates/indexer-receipt/src/lib.rs
suchapalaver: I'm asking you, why you think this won't work:

```rust
pub fn get_receipt(&self) -> &tap_graph::v2::SignedReceipt {
        match self {
            TapReceipt::V2(receipt) => receipt,
            TapReceipt::V1(receipt) => receipt,
        }
    }
```

Maybe I'm missing something!

gusinacio: Each one has a different type, the compiler would say `tap_graph::SignedReceipt` is not an instance of `tap_graph::v2::SignedReceipt`

suchapalaver: Right, I see that now, the `v1` having no similar label threw me. 

But do the `SignedReceipt` types not implement some shared behavior? But it doesn't have to look so confusing?

PR #595, opened 2025-01-31
refactor: store horizon receipts

gusinacio:


crates/service/src/tap/receipt_store.rs
```diff
@@ -17,7 +18,25 @@ pub struct InnerContext {
...
+            self.store_receipts_v1(v1_receipts),
+            self.store_receipts_v2(v2_receipts)
+        );
+        if let Err(e) = insert_v1 {
+            tracing::error!("Failed to store v1 receipts: {}", e);
+        }
+        if let Err(e) = insert_v2 {
+            tracing::error!("Failed to store v2 receipts: {}", e);
+        }
+    }
```
suchapalaver, line 37:
```suggestion
    async fn process_db_receipts(&self, buffer: Vec<DatabaseReceipt>) {
        let (v1_receipts, v2_receipts): (Vec<_>, Vec<_>) =
            buffer.into_iter().partition_map(|r| match r {
                DatabaseReceipt::V1(db_receipt_v1) => Either::Left(db_receipt_v1),
                DatabaseReceipt::V2(db_receipt_v2) => Either::Right(db_receipt_v2),
            });
        let (insert_v1, insert_v2) = tokio::join!(
            self.store_receipts_v1(v1_receipts),
            self.store_receipts_v2(v2_receipts)
        );
        if let Err(e) = insert_v1 {
            tracing::error!("Failed to store v1 receipts: {}", e);
        }
        if let Err(e) = insert_v2 {
            tracing::error!("Failed to store v2 receipts: {}", e);
        }
    }
```

I like what you've done here, but let's return the error and have v1/v2 error variants, no?

crates/service/src/tap/receipt_store.rs
```diff
@@ -81,9 +165,7 @@ impl IndexerTapContext {
                 tokio::select! {
                     biased;
                     _ = receiver.recv_many(&mut buffer, BUFFER_SIZE) => {
-                        if let Err(e) = inner_context.store_receipts(buffer).await {
-                            tracing::error!("Failed to store receipts: {}", e);
-                        }
```
suchapalaver, line 86: I know what I'm suggesting above means bringing this back in, but it's easier to read (and more adaptable) to have an error return here.

suchapalaver (changes requested):
@gusinacio just one suggestion to make `process_db_receipts` and this looks great 👍

migrations/20250131122241_tap_horizon_receipts.down.sql
```diff
@@ -0,0 +1,2 @@
+-- Add down migration script here
+DROP TABLE IF EXISTS scalar_tap_receipts CASCADE;
```
carlosvdr, line 2: Should we also drop in here `tap_horizon_receipts`?

suchapalaver: @gusinacio is there a test we can add to @carlosvdr 's point?

crates/service/src/tap/receipt_store.rs
gusinacio: Should I return both errors? They are different things, one could process and the other could fail.

migrations/20250131122241_tap_horizon_receipts.down.sql
gusinacio: You are correct! Fixed it.

crates/service/src/tap/receipt_store.rs
suchapalaver: Not sure about "should". I would imagine three enum/thiserror variants - with one of them being in the offchance of them both failing?

I guess the concern is that, as you say, all that could happen and the caller of the function could be absolutely none the wiser.

gusinacio: tbh, I just moved this to a function because format and lsp are a mess within macros. The only caller for this function is inside the select!{} macro.

gusinacio: done

crates/service/src/tap/receipt_store.rs
gusinacio: done

migrations/20250131122241_tap_horizon_receipts.down.sql
suchapalaver: Not requiring you add testing (perish the thought!) to this PR, but what would be a testing strategy so that what @carlosvdr caught isn't being left to @carlosvdr catching it each time?

carlosvdr (approved):

suchapalaver (approved):

migrations/20250131122241_tap_horizon_receipts.down.sql
gusinacio: It's only used for test and sqlx. https://github.com/graphprotocol/indexer-rs/blob/main/migrations/README.md

PR #597, opened 2025-01-31
refactor: change the interfaces and struct definitions for dips

pcarranzav:
- Changed the database table for dips to indexing_agreements
- Changed some field names for consistency (camelCase)
- Added pricePerEntity to dips vouchers
- Changed the types used for network names and addresses
- Added a current allocation id to indexing agreements (will be used by the indexer agent)


crates/dips/src/lib.rs
```diff
@@ -479,7 +483,7 @@ mod test {
                 payee: payee_addr,
                 service: Address(FixedBytes::ZERO),
                 metadata: metadata.eip712_hash_struct().to_owned().into(),
-                cancellled_by: signer.address(),
+                cancelledBy: signer.address(),
```
suchapalaver, line 486: This is still British English spelling, "canceled" in US spelling fwiw fyi,

pcarranzav: Good point, I never remember which is which... I see other references to "cancelled" in the repo though (it seems tokio uses cancelled in cancellation tokens https://docs.rs/tokio-util/latest/tokio_util/sync/struct.CancellationToken.html which we use here), so I'd rather leave this as it is for consistency

suchapalaver: TIL on the CancellationToken! As long as we're being consistent then lol

pcarranzav:
Back to Draft as I'm adding a few more changes

mangas (approved):

gusinacio (approved):
LGTM

LNSD (commented):
LGTM ✅

PR #600, opened 2025-02-01
fix: check allocation id blocked before sending rav req

### LINKED ISSUE #598: Error messages while aggregate a closing allocation
## Problem
When you have a lot of receipts for older allocations that were already being closed, once it starts an allocation, it sends a `SenderAccountMessage::UpdateReceiptFees` to let the sender account know about the updated value.

As soon as it gets the message, it verifies that the count is bigger than the limit, and triggers a rav request for that specific allocation. While trying to send a request to the allocation, it was already closed, and no allocation actor was found, giving this message

```
ERROR indexer_tap_agent::agent::sender_account: There was an error while requesting a RAV., error: Error while getting allocation actor 0x....
```

## Solution
We have a flag called "blocked_allocation" which is added when sender account closes an allocation.

If the allocation is blocked (closing), we shouldn't trigger an allocation for `counter_greater_receipt_limit`.

shiyasmohd:
Fixes #598

### DIFF
--- a/crates/tap-agent/src/tracker/generic_tracker.rs
+++ b/crates/tap-agent/src/tracker/generic_tracker.rs
@@ -214,6 +214,13 @@ where
         });
     }
 
+    pub fn is_allocation_id_blocked(&self, address: &Address) -> bool {
+        self.id_to_fee
+            .get(address)
+            .map(|v| v.blocked)
+            .expect("Allocation ID not found")
+    }
+
     pub fn can_trigger_rav(&self, allocation_id: Address) -> bool {
         self.id_to_fee
             .get(&allocation_id)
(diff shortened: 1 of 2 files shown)


crates/tap-agent/src/tracker/generic_tracker.rs
```diff
@@ -214,6 +214,13 @@ where
         });
     }
 
+    pub fn is_allocation_id_blocked(&self, address: &Address) -> bool {
+        self.id_to_fee
+            .get(address)
+            .map(|v| v.blocked)
+            .expect("Allocation ID not found")
```
suchapalaver, line 221: I get it's probably unlikely but maybe let's handle the address not being found with a `match` or something and log it so we don't panic.

suchapalaver:
@shiyasmohd looks like you need to rebase! Also, this all looks good to me, but I wanted to ask as I learn the codebase, would there be a test we could add that shows this fix working? Just generally really curious how people are testing whether it's locally or however. Thanks 🙏

shiyasmohd:
@suchapalaver I'll rebase and try to add test too. Please Ingore this Merge from `main`  branch. I'll be fixing that too. Thanks for the review!

PR #608, opened 2025-02-07
refactor: changes to dips to better support payment collection in indexer-agent

pcarranzav:
- add last_allocation_id and last_payment_collected_at to the indexing_agreements table
- improve the dips store to allow getting all the fields from the db (not needed now, but leaving it anyways as it's a nicer interface if we need it later)
- move the dips database store to the dips crate
- rename the grpc services to IndexerDipsService (indexer side) and DipperService (gateway side)
- configure the tonic build to build both server and client for each service, and bump the npm package accordingly


crates/dips/proto/gateway.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package graphprotocol.gateway.dips;
 
-service DipsService {
```
LNSD, line 5: I would suggest naming it `GatewayDipsService.`

I want to rename the "Dipper" to DIPs Gateway at some point in the future.

crates/dips/proto/indexer.proto
```diff
@@ -2,7 +2,7 @@ syntax = "proto3";
 
 package graphprotocol.indexer.dips;
 
-service DipsService {
+service IndexerDipsService {
```
LNSD, line 5: I would suggest naming it `GatewayDipsService.`

```suggestion
service GatewayDipsService {
```
I want to rename the "Dipper" to DIPs Gateway at some point in the future.

LNSD (approved):
LGTM ✅

crates/service/src/database/mod.rs
```diff
@@ -2,7 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0
 
 pub mod cost_model;
-pub mod dips;
```
gusinacio, line 5: You removed dips from the modules but didn't remove `dips.rs` file. Is there any reason?

pcarranzav: I moved it to database.rs on the dips crate, so it shows in git as moved rather than removed

gusinacio: You are right! Github tricked me

gusinacio (approved):
This PR affects directly DIPs with no changes in the rest of the application. LGTM

PR #609, opened 2025-02-07
refactor: implement receipt traits for horizon

gusinacio:
- Simply filling the gaps that was missing with unimplemented!()
- Using rstest to provide different fixtures with different contexts to the same test

### DIFF
--- a/crates/tap-agent/src/tap/context/receipt.rs
+++ b/crates/tap-agent/src/tap/context/receipt.rs
@@ -208,10 +208,132 @@ impl ReceiptRead<TapReceipt> for TapAgentContext<Horizon> {
 
     async fn retrieve_receipts_in_timestamp_range<R: RangeBounds<u64> + Send>(
         &self,
-        _timestamp_range_ns: R,
-        _receipts_limit: Option<u64>,
+        timestamp_range_ns: R,
+        receipts_limit: Option<u64>,
     ) -> Result<Vec<CheckingReceipt>, Self::AdapterError> {
-        unimplemented!()
+        let receipts_limit = receipts_limit.map_or(1000, |limit| limit);
+
+        let signers = signers_trimmed(self.escrow_accounts.clone(), self.sender)
+            .await
+            .map_err(|e| AdapterError::ReceiptRead {
+                error: format!("{:?}.", e),
+            })?;
+
+        // TODO filter by data_service when we have multiple data services
+
+        let records = sqlx::query!(
+            r#"
+                SELECT 
+                    id,
+                    signature,
+                    allocation_id,
+                    payer,
+                    data_service,
+                    service_provider,
+                    timestamp_ns,
+                    nonce,
+                    value
+                FROM tap_horizon_receipts
+                WHERE
+                    allocation_id = $1
+                    AND payer = $2
+                    AND service_provider = $3
+                    AND signer_address IN (SELECT unnest($4::text[]))
+                AND $5::numrange @> timestamp_ns
+                ORDER BY timestamp_ns ASC
+                LIMIT $6
+            "#,
+            self.allocation_id.encode_hex(),
+            self.sender.encode_hex(),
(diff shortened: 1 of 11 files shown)


carlosvdr (dismissed):
LGTM

crates/tap-agent/src/tap/context/receipt.rs
```diff
@@ -208,10 +208,132 @@ impl ReceiptRead<TapReceipt> for TapAgentContext<Horizon> {
...
+            rangebounds_to_pgrange(timestamp_range_ns),
+            (receipts_limit + 1) as i64,
+        )
+        .fetch_all(&self.pgpool)
+        .await?;
+        let mut receipts = records
+            .into_iter()
+            .map(|record| {
+                let signature = record.signature.as_slice().try_into()
+                    .map_err(|e| AdapterError::ReceiptRead {
```
suchapalaver, line 259: I would love to see these errors being handled via their types instead converting them all to `String`. If you don't want to fix it now, we could create a ticket for that as a separate piece of work.

crates/tap-agent/src/tap/context/receipt.rs
```diff
@@ -208,10 +208,132 @@ impl ReceiptRead<TapReceipt> for TapAgentContext<Horizon> {
...
+                    .to_u64()
+                    .ok_or(AdapterError::ReceiptRead {
+                        error: "Error decoding timestamp_ns while retrieving receipt from database"
+                            .to_string(),
+                    })?;
+                let nonce = record.nonce.to_u64().ok_or(AdapterError::ReceiptRead {
+                    error: "Error decoding nonce while retrieving receipt from database".to_string(),
+                })?;
+                // Beware, BigDecimal::to_u128() actually uses to_u64() under the hood...
+                // So we're converting to BigInt to get a proper implementation of to_u128().
```
suchapalaver, line 311: Awesome note, thanks!

crates/tap-agent/src/tap/context/receipt.rs
gusinacio: I'd rather create a ticket for that. These decisions were carried over from previous implementation. It was supposed to be some error from `tap_core` that couldn't handle anyhow (because of Clone requirements) and they decided to put String.

crates/tap-agent/src/test.rs
```diff
@@ -260,6 +263,65 @@ pub fn create_rav(
     .unwrap()
 }
 
+pub trait CreateReceipt {
```
suchapalaver, line 266: So this is a static function effectively? I confused looking a trait implementation with no `self`.

gusinacio: Yes, we don't need to get anything from Self. We just want different function implementations given the T being implemented

crates/tap-agent/src/tap/context/receipt.rs
```diff
@@ -259,40 +406,74 @@ mod test {
...
+            INDEXER.1,
             SENDER.1,
-            escrow_accounts.clone(),
-        );
+            escrow_accounts,
+        )
+    }
 
+    /// Insert a single receipt and retrieve it from the database using the adapter.
+    /// The point here it to test the deserialization of large numbers.
```
suchapalaver, line 453: This is such a useful comment btw! 🙏

crates/tap-agent/src/test.rs
```diff
@@ -260,6 +263,69 @@ pub fn create_rav(
     .unwrap()
 }
 
+/// Generic implementation of create_received_receipt
+pub trait CreateReceipt {
+    /// This might seem weird at first glance since [Horizon] and [Legacy] implementation have the same
+    /// function signature and don't require &self. The reason is that we can not match over T to get
+    /// all variants because T is a trait and not an enum.
```
suchapalaver, line 270: Awesome, thank you! 🙏

suchapalaver (approved):

PR #621, opened 2025-02-11
feat: add escape hatch to trusted senders

gusinacio:
Add feature for configuration that allows some senders to spend up to `max_amount_willing_to_lose` over the balance.

### DIFF
--- a/crates/config/maximal-config-example.toml
+++ b/crates/config/maximal-config-example.toml
@@ -124,6 +124,11 @@ max_receipt_value_grt = "0.001" # 0.001 GRT. We use strings to prevent rounding 
 # max_amount_willing_to_lose_grt = "0.1"
 max_amount_willing_to_lose_grt = 20
 
+# List of Senders that are allowed to spend up to `max_amount_willing_to_lose_grt`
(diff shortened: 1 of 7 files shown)


crates/config/maximal-config-example.toml
```diff
@@ -124,6 +124,11 @@ max_receipt_value_grt = "0.001" # 0.001 GRT. We use strings to prevent rounding
 # max_amount_willing_to_lose_grt = "0.1"
 max_amount_willing_to_lose_grt = 20
 
+# List of Gateways that are allowed to spend up to `max_amount_willing_to_lose_grt`
+# over the escrow balance
+trusted_gateways = ["0xdeadbeefcafebabedeadbeefcafebabedeadbeef"]
```
Theodus, line 129: Can we use the term "sender", rather than introducing the term "gateway" unnecessarily?

gusinacio: good catch!

suchapalaver (approved):

Theodus (approved):
LGTM

PR #624, opened 2025-02-11
refactor: create senders based on escrow account

gusinacio:


crates/tap-agent/src/agent/sender_accounts_manager.rs
```diff
@@ -500,12 +615,18 @@ impl State {
...
-        // TODO Load horizon allocations
-
         unfinalized_sender_allocations_map
     }
 
+    /// Gather all outstanding receipts and unfinalized RAVs from the database.
+    /// Used to create [SenderAccount] instances for all senders that have unfinalized allocations
+    /// and try to finalize them if they have become ineligible.
+    ///
+    /// This loads horiazon allocations
```
gusinacio, line 625: typo

crates/monitor/src/escrow_accounts.rs
```diff
@@ -95,12 +95,33 @@ pub async fn escrow_accounts(
     reject_thawing_signers: bool,
 ) -> Result<EscrowAccountsWatcher, anyhow::Error> {
     indexer_watcher::new_watcher(interval, move || {
-        get_escrow_accounts(escrow_subgraph, indexer_address, reject_thawing_signers)
+        get_escrow_accounts_v1(escrow_subgraph, indexer_address, reject_thawing_signers)
     })
     .await
 }
```
carlosvdr, line 101: I think this function should be renamed to escrow_accounts_v1, it can make it a bit confusing? since in every other spot you are making the difference in the names between v1 and v2 and this could give the idea is generic for both versions

crates/tap-agent/src/agent.rs
```diff
@@ -154,7 +156,16 @@ pub async fn start_agent() -> (ActorRef<SenderAccountsManagerMessage>, JoinHandl
...
 
-    let escrow_accounts = escrow_accounts(
+    let escrow_accounts_v1 = escrow_accounts(
+        escrow_subgraph,
+        *indexer_address,
+        *escrow_sync_interval,
+        false,
+    )
+    .await
+    .expect("Error creating escrow_accounts channel");
```
suchapalaver, line 166: Do we want to panic here?

crates/monitor/src/escrow_accounts.rs
```diff
@@ -95,12 +95,33 @@ pub async fn escrow_accounts(
...
+
+// TODO implement escrow accounts v2 query
+async fn get_escrow_accounts_v2(
+    _escrow_subgraph: &'static SubgraphClient,
+    _indexer_address: Address,
+    _reject_thawing_signers: bool,
+) -> anyhow::Result<EscrowAccounts> {
+    Ok(EscrowAccounts::new(HashMap::new(), HashMap::new()))
+}
+
```
carlosvdr, line 123: Add TODO: Change query for temporary implementation of subgraph handling both v1 and V2

justfile
```diff
@@ -14,8 +14,15 @@ url:
 clippy:
     cargo +nightly clippy --all-targets --all-features
 
+ci:
+  just fmt
+  just clippy
+  just test
+  just sqlx-prepare
```
suchapalaver, line 21: Could you add a doc comment on this this please @gusinacio 🙏

crates/tap-agent/tests/sender_account_manager_test.rs
```diff
@@ -124,19 +142,20 @@ async fn sender_account_manager_layer_test(pgpool: PgPool) {
...
     // verify if it gets removed
-    let actor_ref = ActorRef::<SenderAccountMessage>::where_is(format!("{}:{}", prefix, SENDER.1));
+    let actor_ref =
+        ActorRef::<SenderAccountMessage>::where_is(format!("{}:legacy:{}", prefix, SENDER.1));
     assert!(actor_ref.is_none());
 
     let rav_marked_as_last = sqlx::query!(
         r#"
-            SELECT * FROM scalar_tap_ravs WHERE last;
+            SELECT * FROM scalar_tap_ravs;
```
carlosvdr, line 158: Why was `WHERE last` removed?

gusinacio: Good catch! I missed to add it back after debugging.

crates/monitor/src/escrow_accounts.rs
gusinacio: There's already a TODO on top of the function. Do you want me to change the messages?

suchapalaver (approved):

suchapalaver (approved):

suchapalaver:
Nice one @gusinacio taking inspiration from #625 👍

crates/monitor/src/escrow_accounts.rs
carlosvdr: Maybe just create an issue so that this is not forgotten?

crates/monitor/src/escrow_accounts.rs
carlosvdr: I just tried to mean that the current comment kind of only sounds for v2 and we might need to modify also v1 query later

carlosvdr (approved):

crates/monitor/src/escrow_accounts.rs
gusinacio: At the moment you change the provided schema old queries will break so you will quickly realize that you need to change the query.

On the other hand, this won't break and that's why there's a TODO

PR #625, opened 2025-02-12
perf: speed up startup setup queries

rotarur:
This PR refactor the queries for scalar_tap_ravs and scalar_tap_receipts to optimize the resources and time
when running in a database with many receipts.

What changed:

refactored the queries for optimization


gusinacio (approved):
LGTM.

PR #626, opened 2025-02-12
refactor: use correct db tables in sender allocation

gusinacio:
Create a trait called `DatabaseInteractions`.

All calls to databases are delegated to trait like:
- mark_rav_last
- calculate_fee_until_last_id
- delete_receipts_between

store_invalid_receipts is implemented in two functions and we use `partition_map` to provide the correct Vector for each


suchapalaver (approved):
Looks real good to me

carlosvdr (approved):

PR #628, opened 2025-02-12
refactor: listen to v2 receipts

gusinacio:
- Enable listening to horizon receipts, they should have the same structure since receipts are minimal and we don't need to carry all fields.


suchapalaver (dismissed):
👍

carlosvdr (dismissed):

suchapalaver (approved):
LGTM

PR #631, opened 2025-02-13
refactor: parse v2 receipts

gusinacio:
- Try to decode a base64 before trying serde_json.
- If success, proceed to decode using Protobuf for v2
- otherwise use normal v1 serde_json deserialization process


suchapalaver (approved):
❤️

carlosvdr (approved):

PR #632, opened 2025-02-14
test: add rstest and fixtures

carlosvdr:
Added `rstest`:

- with this several tests have been refactored to only receive fixtures created on `rstest`

this PR aims to make the tests more easy reading and only focusing on the important bits per test

Due to the nature of async fixtures some tests could use them since the test itself needed to be in some order for it to work as intended.

### DIFF
--- a/crates/monitor/src/client/monitor.rs
+++ b/crates/monitor/src/client/monitor.rs
@@ -72,8 +72,14 @@ mod tests {
 
     use super::*;
 
-    #[tokio::test]
-    async fn test_parses_synced_and_healthy_response() {
+    struct MonitorMock {
+        mock_server: MockServer,
+        status_url: Url,
+        deployment: DeploymentId,
+    }
+
+    #[rstest::fixture]
+    async fn monitor_mock() -> MonitorMock {
         let mock_server = MockServer::start().await;
         let status_url: Url = mock_server
             .uri()
@@ -82,7 +88,16 @@ mod tests {
             .join("/status")
             .unwrap();
         let deployment = deployment_id!("QmAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA");
+        MonitorMock {
+            mock_server,
+            status_url,
+            deployment,
+        }
+    }
 
+    #[rstest::rstest]
+    #[tokio::test]
+    async fn test_parses_synced_and_healthy_response(#[future(awt)] monitor_mock: MonitorMock) {
         Mock::given(method("POST"))
             .and(path("/status"))
             .respond_with(ResponseTemplate::new(200).set_body_json(json!({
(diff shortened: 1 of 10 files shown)


crates/service/Cargo.toml
```diff
@@ -74,6 +74,7 @@ tokio-test = "0.4.4"
 wiremock.workspace = true
 insta = "1.41.1"
 test-log.workspace = true
+async-std = { version = "1.12.0", features = ["attributes"] }
```
TypeLevelConsoli, line 78: I don't think `async-std` is necessary, we're using `#[tokio::test]` on every async test case (same thing for `tap-agent`)

crates/tap-agent/src/agent/sender_account.rs
```diff
@@ -1467,6 +1468,22 @@ pub mod tests {
                 .await;
         mock_escrow_subgraph_server
     }
+    struct TestSenderAccount {
```
TypeLevelConsoli, line 1471: I like that you created more structured outputs for the fixtures!

TypeLevelConsoli (changes requested):
fixtures indeed made the code clearer, thanks for this

crates/monitor/src/client/monitor.rs
```diff
@@ -82,7 +88,16 @@ mod tests {
...
+            mock_server,
+            status_url,
+            deployment,
+        }
+    }
 
+    #[rstest::rstest]
+    #[tokio::test]
+    async fn test_parses_synced_and_healthy_response(#[future(awt)] monitor_mock: MonitorMock) {
         Mock::given(method("POST"))
```
neithanmo, line 101: I noticed that `Mock::given` is used across multiple tests. This pattern seems like a good candidate for a fixture to reduce duplication and improve test clarity???.

According to [this](https://docs.rs/rstest/0.25.0/rstest/#creating-a-test-for-each-combinations-of-given-values), You could leverage rstest's parameterized fixtures like bellow(pseudo-code):

```rust
#[rstest::fixture]
async fn status_mock(
    #[await] monitor_mock: MonitorMock, 
    #[values(true, false)] synced: bool,  // <------ this is a parameterized fixture argument
    #[values("healthy", "unhealthy", "failed")] health: &str // <---- this gets also parameterized
) -> (// return type);
```
as synced and health flags are what change the most in some tests. 
But it is nice the other well-structured fixtures you added, are really good!

carlosvdr: This is pretty cool, thanks just tried it out, simplifies this a lot!

neithanmo (approved):
Looks good to me

suchapalaver (approved):

TypeLevelConsoli (approved):
great!

PR #635, opened 2025-02-17
feat: Escrow based signer validation

mangas:


crates/service/src/service.rs
```diff
@@ -135,13 +136,30 @@ pub async fn run() -> anyhow::Result<()> {
         let ipfs_fetcher: Arc<dyn IpfsFetcher> =
             Arc::new(IpfsClient::new("https://api.thegraph.com/ipfs/").unwrap());
 
+        // TODO: Try to re-use the same watcher for both DIPS and TAP
+        let watcher = escrow_accounts_v2(
```
pcarranzav, line 140: worth checking with the Semiotic team but I'm 99% sure we need escrow_accounts_v1, as v2 is the one for after we launch Horizon

anirudh2: That's correct.

pcarranzav (dismissed):

suchapalaver (dismissed):

anirudh2 (dismissed):
Looks great! I'm approving but you mind creating an issue for this TODO so that we don't lose track of it.

`// TODO: Try to re-use the same watcher for both DIPS and TAP`

mangas:
> Looks great! I'm approving but you mind creating an issue for this TODO so that we don't lose track of it.
> 
> `// TODO: Try to re-use the same watcher for both DIPS and TAP`

Added a ticket to the board

pcarranzav (approved):

PR #643, opened 2025-02-26
refactor: use educe instead of manually implementing traits for tests

TypeLevelConsoli:
Remove the manual implementation of PartialEq/Eq for tests to use the [`educe`](https://docs.rs/educe/latest/educe/) crate.
This makes the code well organized.

### DIFF
--- a/crates/tap-agent/src/agent/sender_account.rs
+++ b/crates/tap-agent/src/agent/sender_account.rs
@@ -102,6 +102,7 @@ type Balance = U256;
 
 /// Information for Ravs that are abstracted away from the SignedRav itself
 #[derive(Debug, Default, PartialEq, Eq)]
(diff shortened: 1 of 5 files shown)


crates/tap-agent/src/agent/sender_account.rs
```diff
@@ -1415,50 +1443,6 @@ pub mod tests {
...
-                    core::mem::discriminant(self),
-                    core::mem::discriminant(other),
-                ) {
-                    (a, b) if a != b => false,
-                    _ => unimplemented!("PartialEq not implementated for {a:?} and {b:?}"),
-                },
-            }
-        }
-    }
-
```
suchapalaver, line 1461: Lovely stuff!

suchapalaver (approved):
Thanks for the nice lesson in how to use `educe`!

PR #647, opened 2025-03-06
refactor: use channels on TestableActor to be able to verify sent messages

TypeLevelConsoli:
TestableActors uses an `Arc<Notify>` to know when their messages are triggered, but that only notifies a mute signal, no context or message is gathered. By changing it to `mpsc::channel` instead, we can receive the message passed and compare it to that we sent.


suchapalaver (approved):
Really clear solution, I appreciated how this made the tests themselves more readable 👍

PR #652, opened 2025-03-13
Fix SubgraphServiceError to return JSON-formatted error responses

neithanmo:
Updates the SubgraphServiceError implementation to return proper JSON-formatted responses, consistent with IndexerServiceError. This fixes JSON parsing errors in GraphQL clients when handling error cases, for example introspection queries

### DIFF
--- a/crates/service/src/error.rs
+++ b/crates/service/src/error.rs
@@ -37,6 +37,25 @@ pub enum IndexerServiceError {
     EscrowAccount(#[from] EscrowAccountsError),
 }
 
+// Helper struct to properly format
+// error messages
+#[derive(Serialize)]
+struct ErrorResponse {
+    message: String,
+}
+
+impl ErrorResponse {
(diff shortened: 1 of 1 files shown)


suchapalaver (approved):

crates/service/src/error.rs
```diff
@@ -37,6 +37,25 @@ pub enum IndexerServiceError {
...
+    fn new(message: impl ToString) -> Self {
+        Self {
+            message: message.to_string(),
+        }
+    }
+
+    fn into_response(self, status_code: StatusCode) -> Response {
+        (status_code, Json(self)).into_response()
+    }
+}
```
TypeLevelConsoli, line 57: nice, structured errors are always a good thing to have! :)

TypeLevelConsoli (approved):

carlosvdr (approved):

PR #653, opened 2025-03-13
fix: skip processing of receipts when there's no receipt for a version

TypeLevelConsoli:
Skip the processing of receipts of a version, when there's no receipt for that version.

This should fix our release issue, as we won't try to access horizon tables when processing v1 receipts anymore.

### DIFF
--- a/crates/service/Cargo.toml
+++ b/crates/service/Cargo.toml
@@ -63,6 +63,7 @@ pin-project = "1.1.7"
 tonic.workspace = true
 itertools = "0.14.0"
 
+
 [dev-dependencies]
 hex-literal = "0.4.1"
 test-assets = { path = "../test-assets" }
@@ -74,6 +75,7 @@ tokio-test = "0.4.4"
 wiremock.workspace = true
 insta = "1.41.1"
 test-log.workspace = true
+futures = "0.3.31"
 
 [build-dependencies]
 build-info-build = { version = "0.0.40", default-features = false }
--- a/crates/service/src/tap/receipt_store.rs
+++ b/crates/service/src/tap/receipt_store.rs
@@ -30,7 +30,7 @@ enum ProcessReceiptError {
 /// Indicates which versions of Receipts where processed
 /// It's intended to be used for migration tests
 #[derive(Debug, PartialEq, Eq)]
-pub enum Processed {
+pub enum ProcessedReceipt {
     V1,
     V2,
     Both,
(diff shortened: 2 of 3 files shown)


crates/service/src/tap/receipt_store.rs
```diff
@@ -27,25 +27,50 @@ enum ProcessReceiptError {
     Both(anyhow::Error, anyhow::Error),
 }
 
+/// Indicates which versions of Receipts where processed
+/// It's intended to be used for migration tests
+#[derive(Debug, PartialEq, Eq)]
+pub enum Processed {
+    V1,
+    V2,
+    All,
```
neithanmo, line 36: I think this type is useful beyond testing!
The "All" variant is a bit confusing to me tho, does it mean every receipt was processed, or that receipts from both versions were processed?
Maybe "Both"  is better for clarity??

TypeLevelConsoli: I agree! "Both" was my first choice, and I changed for no reason. I changed it back to `Both`

carlosvdr:
This is amazing, and a great wait to test migrations!

crates/service/src/tap/receipt_store.rs
```diff
@@ -27,25 +27,50 @@ enum ProcessReceiptError {
...
             buffer.into_iter().partition_map(|r| match r {
                 DatabaseReceipt::V1(db_receipt_v1) => Either::Left(db_receipt_v1),
                 DatabaseReceipt::V2(db_receipt_v2) => Either::Right(db_receipt_v2),
             });
-        let (insert_v1, insert_v2) = tokio::join!(
-            self.store_receipts_v1(v1_receipts),
-            self.store_receipts_v2(v2_receipts)
-        );
+
+        let (insert_v1, insert_v2) = match (v1_receipts.is_empty(), v2_receipts.is_empty()) {
```
neithanmo, line 51: Just thinking, what if we modify the helper store functions instead?
```rust
async fn store_receipts_v1(&self, receipts: Vec<DbReceiptV1>) -> Result<bool, AdapterError> {
    if receipts.is_empty() {
        return Ok(false);
    }
    
    // existing code
    
    Ok(true) // > this means actually data was stored
}
```
The change lets the helper functions handle the "was anything processed?" logic. Then we could simplify the main function by removing the empty-check and keep returning the Processed type as it is now.

It follows your original idea,  what do you think??

TypeLevelConsoli: absolutely! This idea made the code much easier to reason about, thanks!
I made the functions return the affected rows instead of a boolean because I think it fits better, but the result is the same.

neithanmo: indeed, returning the number of affected rows is much better !!!

neithanmo (approved):

crates/service/Cargo.toml
```diff
@@ -63,6 +63,7 @@ pin-project = "1.1.7"
 tonic.workspace = true
 itertools = "0.14.0"
 
+
```
suchapalaver, line 66: Is that a style thing?

crates/service/src/tap/receipt_store.rs
```diff
@@ -27,29 +27,49 @@ enum ProcessReceiptError {
     Both(anyhow::Error, anyhow::Error),
 }
 
+/// Indicates which versions of Receipts where processed
+/// It's intended to be used for migration tests
```
suchapalaver, line 31: Really appreciate the docs!

suchapalaver (dismissed):
I love the testing strategy here using the `Migrator` and I think this means we can consider releasing again now 🚀

crates/service/Cargo.toml
TypeLevelConsoli: yep, but from cargo
I just `cargo add futures --dev` and it also added this line :man_shrugging:

TypeLevelConsoli:
I rebased the branch and `cargo` reordered a line from `Cargo.lock`

suchapalaver (approved):

neithanmo (approved):

PR #660, opened 2025-03-20
Enhance middleware error reporting and logging without breaking changes

### LINKED ISSUE #619: middleware errors not logged
I have encountered situations that are unnecessarily difficult to debug because an error result from a middleware function is neither reported in the response body nor logged via tracing. For example, when the `attestation_middleware` returns `AttestationError::CouldNotFindSigner` the response just has status 500 with no body explaining what went wrong and there are no logs indicating an issue:
```
  2025-02-11T15:43:29.023335Z  INFO indexer_service_rs::metrics: Serving prometheus metrics, address: 0.0.0.0:7300
    at crates/service/src/metrics.rs:39

  2025-02-11T15:43:29.023364Z  INFO indexer_service_rs::service: Serving requests, address: 0.0.0.0:7601
    at crates/service/src/service.rs:123
```

Only after building the indexer-service from source with additional logging was I able to discover where in the pile of middleware something went wrong. There should be logs for middleware errors.

neithanmo:
This PR improves error reporting and logging in our middleware components(for attestation) without changing their existing behavior. It addresses issues where errors in middleware functions were neither reported in response bodies(just an error code) nor logged via tracing, making debugging unnecessarily difficult.

Key Changes
- Added structured error responses using `ErrorResponse` for better client feedback
- Implemented detailed tracing logs for error conditions, however, the current interface does not treat them as errors but as optional state changes. 
- Maintained backward compatibility with the existing middleware chain behavior
- Added explicit error/warning messages when signers are missing but attestation is requested

Fixes #619 - Error results from middleware functions not reported in logs or responses

### DIFF
--- a/crates/service/src/middleware/attestation.rs
+++ b/crates/service/src/middleware/attestation.rs
@@ -98,7 +98,7 @@ impl StatusCodeExt for AttestationError {
 
 impl IntoResponse for AttestationError {
     fn into_response(self) -> Response {
-        tracing::error!(error=self, "Attestation error");
+        tracing::error!(error=%self, "Attestation error");
         let status_code = self.status_code();
         ErrorResponse::new(self).into_response(status_code)
     }
--- a/crates/service/src/middleware/attestation.rs
+++ b/crates/service/src/middleware/attestation.rs
@@ -14,7 +14,7 @@ use reqwest::StatusCode;
 use serde::Serialize;
 use thegraph_core::attestation::Attestation;
 
-use crate::error::StatusCodeExt;
+use crate::error::{ErrorResponse, StatusCodeExt};
 
 #[derive(Clone)]
 pub enum AttestationInput {
@@ -56,6 +56,11 @@ pub async fn attestation_middleware(
         (Some(signer), Some(AttestationInput::Attestable { req })) => {
             Some(signer.create_attestation(req, &res))
         }
+        (None, Some(AttestationInput::Attestable { .. })) => {
+            // keep this branch separated just to log the missing signer condition
+            tracing::warn!("Attestation requested but no signer found");
+            None
+        }
         _ => None,
     };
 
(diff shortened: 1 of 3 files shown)


suchapalaver (approved):
Nice! I take it this came out of the experience of working on testing patterns?

crates/service/src/middleware/attestation.rs
```diff
@@ -93,12 +98,14 @@ impl StatusCodeExt for AttestationError {
 
 impl IntoResponse for AttestationError {
     fn into_response(self) -> Response {
-        self.status_code().into_response()
+        tracing::error!("Attestation error: {}", self);
```
TypeLevelConsoli, line 101:
```suggestion
        tracing::error!(error=self, "Attestation error");
```

TypeLevelConsoli (changes requested):
great, I'm all in for structured errors

crates/service/src/middleware/attestation_signer.rs
```diff
@@ -30,6 +30,12 @@ pub async fn signer_middleware(
     if let Some(Allocation(allocation_id)) = request.extensions().get::<Allocation>() {
         if let Some(signer) = state.attestation_signers.borrow().get(allocation_id) {
             request.extensions_mut().insert(signer.clone());
+        } else {
+            // Just log this case which is silently passed through next middleware
+            tracing::warn!(
+                "No attestation signer found for allocation {}",
+                allocation_id
+            );
```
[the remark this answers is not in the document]
TypeLevelConsoli: for example, the current code will create a tracing event such as:

`2025-03-20T21:04:49.336978Z  WARN foo: No attestation signer found for allocation 123`

while structured params will give us:

`2025-03-20T21:05:10.407913Z  WARN foo: No attestation signer found for allocation allocation_id=123`

and so we can filter for both the message or a specific allocation_id

neithanmo:
> great, I'm all in for structured errors

Thanks! @TypeLevelConsoli for suggestion, indeed, that way is clearer

crates/service/src/middleware/attestation_signer.rs
```diff
@@ -30,6 +30,11 @@ pub async fn signer_middleware(
     if let Some(Allocation(allocation_id)) = request.extensions().get::<Allocation>() {
         if let Some(signer) = state.attestation_signers.borrow().get(allocation_id) {
             request.extensions_mut().insert(signer.clone());
+        } else {
+            // Just log this case which is silently passed through next middleware
+            tracing::warn!(warn=%allocation_id,
```
TypeLevelConsoli, line 35: having the field named `warn` is not very helpful, instead, we should log it as `allocation_id` for better context

```suggestion
            tracing::warn!(allocation_id,
```

TypeLevelConsoli (changes requested):

TypeLevelConsoli (approved):

PR #666, opened 2025-03-25
fix(dips): deserialize dips message signature from bytes

LNSD:
This pull request includes changes to the `Cargo.toml` and `lib.rs` files in the `crates/dips` directory.

* [`crates/dips/src/lib.rs`](diffhunk://#diff-43049e4db6e400dbf78fe60574626cb162e14dbb774b13f92bc9b05b68d08830L204-R209): Refactored the signature handling in the `SignedIndexingAgreementVoucher` implementation to use `Signature::try_from` to fix the signature deserialization bug.

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -198,7 +198,7 @@ impl SignedIndexingAgreementVoucher {
         expected_payee: &Address,
         allowed_payers: impl AsRef<[Address]>,
     ) -> Result<(), DipsError> {
-        let sig = Signature::from_str(&self.signature.to_string())
+        let sig = Signature::try_from(self.signature.as_ref())
             .map_err(|err| DipsError::InvalidSignature(err.to_string()))?;
 
         let payer = self.voucher.payer;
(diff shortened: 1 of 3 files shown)


crates/dips/src/lib.rs
```diff
@@ -201,8 +201,12 @@ impl SignedIndexingAgreementVoucher {
         expected_payee: &Address,
         allowed_payers: impl AsRef<[Address]>,
     ) -> Result<(), DipsError> {
-        let sig = Signature::from_str(&self.signature.to_string())
-            .map_err(|err| DipsError::InvalidSignature(err.to_string()))?;
+        let sig = match Signature::try_from(self.signature.as_ref()) {
```
neithanmo, line 204: It would be better if you remove the match statement and use `map_err(|e| DipsError::InvalidSignature(e.to_string()))?` instead.

neithanmo (changes requested):

crates/dips/src/lib.rs
LNSD: Done! 📝

pcarranzav (approved):

neithanmo (approved):

PR #667, opened 2025-03-25
fix: use the correct eip712 domain for each part of dips

pcarranzav:


neithanmo (approved):
Thanks!
Looks good to me
@TypeLevelConsoli can you take a look?

PR #671, opened 2025-03-26
fix: return error instead of panicking if fetching from ipfs fails

pcarranzav:


crates/dips/src/ipfs.rs
```diff
@@ -50,7 +50,7 @@ impl IpfsFetcher for IpfsClient {
             .map_ok(|chunk| chunk.to_vec())
             .try_concat()
             .await
-            .unwrap();
+            .map_err(|_| DipsError::SubgraphManifestUnavailable(file.to_string()))?;
```
mangas: could be worth logging the error here so we can see why it failed

mangas (approved):

mangas (approved):

PR #675, opened 2025-03-27
fix: validate dips pricing

mangas:


crates/service/src/service.rs
```diff
@@ -148,13 +149,16 @@ pub async fn run() -> anyhow::Result<()> {
         .await
         .expect("Failed to create escrow accounts watcher");
 
+        let registry = NetworksRegistry::from_latest_version().await.unwrap();
```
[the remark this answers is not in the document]
pcarranzav: Panic is correct in this case, as we wouldn't want to continue starting the service if this fails

crates/config/src/config.rs
```diff
@@ -396,6 +396,10 @@ pub struct DipsConfig {
     pub host: String,
     pub port: String,
     pub allowed_payers: Vec<Address>,
+
+    pub price_per_entity: u64,
+    pub price_per_epoch: BTreeMap<String, u64>,
```
LNSD, line 401: Use `U256` instead of `u64` with any "wei GRT" value. This will avoid any conversions later.

As a rule of thumb: _prefer an early validation and conversion, over a "scattered" conversion_.

crates/dips/src/lib.rs
```diff
@@ -330,19 +350,28 @@ pub async fn validate_and_create_agreement(
...
-                offered_price.to_string(),
+                offered_epoch_price.to_string(),
             ))
         }
         Some(_) => {}
         None => return Err(DipsError::UnsupportedChainId(metadata.chainId)),
     }
 
+    let offered_entity_price = metadata.pricePerEntity;
+    if offered_entity_price < U256::from(price_calculator.entity_price()) {
```
LNSD, line 367: This conversion can be avoided if all the values use the same type: `U256`.

crates/dips/src/lib.rs
```diff
@@ -754,7 +783,17 @@ mod test {
             subgraphDeploymentId: voucher_ctx.deployment_id.clone(),
         };
 
-        let low_price_voucher = voucher_ctx.test_voucher(metadata);
+        let low_entity_price_voucher = voucher_ctx.test_voucher(metadata);
+
+        let metadata = SubgraphIndexingVoucherMetadata {
+            basePricePerEpoch: U256::from(10_u64),
+            pricePerEntity: U256::from(10000_u64),
```
LNSD, line 790: Idem

crates/dips/src/registry.rs
LNSD, line 1: If everything under this file is for testing purposes, I'd instead put the `#[cfg(test)]` on the `mod` declaration.

crates/dips/src/lib.rs
```diff
@@ -18,6 +20,7 @@ pub mod ipfs;
 pub mod price;
 #[cfg(feature = "rpc")]
 pub mod proto;
+pub mod registry;
```
LNSD, line 23: If everything within this file is only for testing purposes, mark it as test only.

```suggestion
#[cfg(test)]
pub mod registry;
```

LNSD, line 23: Is there any reason for this module to be `pub`?

LNSD (changes requested):
Please, check my comments 🙂

crates/dips/src/lib.rs
mangas: Not really, these are hardcoded values for test which require U256

LNSD (approved):
LGTM ✅

PR #685, opened 2025-04-09
Fix/v2 crash

neithanmo:
This PR implements a feature flag mechanism for Horizon support to prevent crashes. Key changes:

- Added `horizon.enabled = false` configuration option in all relevant config files
- Created a new HorizonConfig struct to properly manage the feature flag, and more settings if needed in the future.
- Modified Horizon-related code to check the enabled flag before execution
- Added conditional logic around database operations for Horizon
- Updated tests to support the new configuration parameter
- Renamed test function from tap_rav_test to test_tap_rav_v1 for clarity
- Removed commented database migration code from start scripts

This change allows systems to run without Horizon support until full implementation is complete. The flag is set to false by default for safety.

### DIFF
--- a/contrib/indexer-service/config.toml
+++ b/contrib/indexer-service/config.toml
@@ -44,3 +44,6 @@ trigger_value_divisor = 500_000
 
 [tap.sender_aggregator_endpoints]
 "ACCOUNT0_ADDRESS_PLACEHOLDER" = "http://tap-aggregator:7610"
+
+[horizon]
+enabled = false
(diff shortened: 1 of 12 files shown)


neithanmo:
i run the test locally and renames the test function to make it clear that the rav being tested is V1.
no crashes due to horizon missing tables anymore.

We need for sure to enable this local testnet integration test in our CI

suchapalaver (approved):
Thanks for this @neithanmo! It's really clear 💪

TypeLevelConsoli (approved):
That's wonderful. Clean and direct solution. Thank you so much!

PR #704, opened 2025-04-27
Add Profiling Support for Indexer/tap-agent services

neithanmo:
This PR implements comprehensive profiling support for the indexer service to help identify performance bottlenecks.

## Changes

- Added embedded continuous profiling using pprof with feature flags `profiling`
- Implemented Docker container support for profiling tools
- Added Just commands for easy profiling management
- Support for both embedded profiling and external tools:
  - Embedded pprof (continuous flamegraphs)
  - Valgrind (memory profiling)
  - Callgrind (CPU profiling)
  - Strace (system call tracing)

## Why embedded profiler
- Because local testnet relies on docker containers, some linux profiling tools do not work well when containerized
- cargo flamegraph, ptrace did not work

as a result, embedding the profiler was the best option and it worked out of the box

### Note
Profiler is disabled even during testing(unit and integration rust tests), because of a weird timing issue, it seems that spawning and setting up profiling threads caused issues with the `test_thawing_deposit_process` test. But this is still useful for end to end testing where it works just fine without huge side effects(as long as tests take into account the delay caused by the profiler during service startup). 

Next step is to prepare a test to run services under heavy load and see how they behave  and look for possible bottlenecks.


suchapalaver (approved):
Awesome :)

suchapalaver (approved):

PR #711, opened 2025-05-08
fix(tap): fix unused v2 deny list check

suchapalaver:
This was always placeholder code to map out how the v2 path would look. This fixes an obvious bug in the unused v2 code.

Signed off by Joseph Livesey <joseph@semiotic.ai>


TypeLevelConsoli (approved):
good catch

PR #715, opened 2025-05-10
fix(indexer): propagate database errors back to the caller

neithanmo:
This changes:
- updates local network to latest release
- Use Oneshot channels to ensure any database error is sent back to the caller(duplicated receipts by design do not trigger any error as database table allows that)
- Update unit-test code accordingly to change above
- Add an integration test that sends receipts directly to indexer to check that malformed receipts trigger errors
- some minor refactors in the integration tests and helper function.


integration-tests/src/rav_tests.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2023-, Edge & Node, GraphOps, and Semiotic Labs.
+// Copyright 2023-, Edge i Node, GraphOps, and Semiotic Labs.
```
TypeLevelConsoli, line 1:
```suggestion
// Copyright 2023-, Edge & Node, GraphOps, and Semiotic Labs.
```

TypeLevelConsoli (changes requested):
Great! Thanks for this!
There is a small typo in the Copyright comment

integration-tests/src/rav_tests.rs
neithanmo: good catch, i will fix it!!!
Thanks!!!

TypeLevelConsoli (approved):

suchapalaver (approved):

PR #720, opened 2025-05-14
fix(docker): add libsasl2-2 to images

### LINKED ISSUE #719: Newer images error on missing libsasl2.so.2
**Describe the bug**
Newer images (from service >= v1.5.3 and tap-agent >= 1.9.4) error out with missing libsasl2.so.2.

**To Reproduce**

```
podman run -it ghcr.io/graphprotocol/indexer-service-rs:v1.5.3 --help
```

results in:

```
/usr/local/bin/indexer-service-rs: error while loading shared libraries: libsasl2.so.2: cannot open shared object file: No such file or directory
```

**Expected behavior**
```
Usage: indexer-service-rs [OPTIONS]

Options:
      --config <FILE>  Path to the configuration file.
                       See https://github.com/graphprotocol/indexer-rs/tree/main/service for examples.
  -h, --help           Print help
  -V, --version        Print version
```

suchapalaver:
Fixes #719 

Signed off by Joseph Livesey <joseph@semiotic.ai>

### DIFF
--- a/Dockerfile.indexer-service-rs
+++ b/Dockerfile.indexer-service-rs
@@ -16,7 +16,7 @@ RUN cargo build --release --bin indexer-service-rs
 FROM debian:bookworm-slim
 
 RUN apt-get update && apt-get install -y --no-install-recommends \
-    openssl ca-certificates protobuf-compiler \
+    openssl ca-certificates protobuf-compiler libsasl2-2 \
     && rm -rf /var/lib/apt/lists/*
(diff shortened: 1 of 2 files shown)


neithanmo (approved):
Thanks for this!!

neithanmo:
I think we should unify images, they are not that different.
However the only inconvenience is that the ones in contrib contains more stuffs that final users might not need

suchapalaver:
> I think we should unify images, they are not that different. However the only inconvenience is that the ones in contrib contains more stuffs that final users might not need

I think first step we should put in place is a test that the images build!

PR #724, opened 2025-05-19
Fix gateway initialization by properly funding escrow during setup

neithanmo:
The gateway service was failing to initialize properly because the escrow contract wasn't being funded correctly. This was causing the gateway's health check to fail with connection reset errors, even though the container was running.

- Added proper error handling and logging for the escrow funding process
- Integrated the escrow funding into the main setup script at the correct point (after gateway deployment, before health check)

This also has the benefit of using local-network configuration directly in our funding instead of defining constants.

### Note

Previous funding script remains in place, maybe useful out of the local-testnet setup

### DIFF
--- a/setup-test-network.sh
+++ b/setup-test-network.sh
@@ -29,10 +29,71 @@ echo "Containers size: $START_CONTAINERS_SIZE"
 echo "Volumes size: $START_VOLUMES_SIZE"
 echo "=============================================="
 
-# Your existing script starts here
 container_running() {
     docker ps --format '{{.Names}}' | grep -q "^$1$"
     return $?
+}
+
+# Function to fund the escrow smart contract
+# 1. first read .env variables from local-network/.env
+# 2. then read contract addresses from local-network/contracts.json
+# 3. finally, use the cast command to approve and deposit GRT to the escrow
+# this should be done just after deploying the gateway
+# otherwise it does not move forward in its setup process
+# causing false error during deployment of our local testnet
+fund_escrow() {
+    echo "Funding escrow for sender..."
+
+    if [ -f "local-network/.env" ]; then
+        source local-network/.env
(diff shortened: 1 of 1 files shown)


setup-test-network.sh
```diff
@@ -29,12 +29,73 @@ echo "Containers size: $START_CONTAINERS_SIZE"
...
     return $?
 }
 
+# Function to fund the escrow smart contract
+# 1. first read .env variables from local-network/.env
+# 2. then read contract addresses from local-network/contracts.json
+# 3. finally, use the cast command to approve and deposit GRT to the escrow
+# this should be done just after deploying the gateway
+# otherwise it does not move forward in its setup process
+# causing false error during deployment of our local testnet
```
suchapalaver, line 43: Great docs!

setup-test-network.sh
```diff
@@ -171,10 +232,6 @@ rm docker-compose.override.yml
 timeout 30 bash -c 'until docker ps | grep indexer | grep -q healthy; do sleep 5; done'
 timeout 30 bash -c 'until docker ps | grep tap-agent | grep -q healthy; do sleep 5; done'
 
-# Mine some blocks
-# This is important for the gateway
-(./local-network/scripts/mine-block.sh 10) 2>/dev/null || true
-
```
suchapalaver, line 177: What was the story with taking this away, just out of curiosity?

suchapalaver (approved):
Love how you're learning more and more about the system!

setup-test-network.sh
neithanmo: The previous version of local network required that(based on my experience not really a impl requirement) to "register" the smart contract deployment and indexer for the gateway.
Now it seems it is enough with the funding just before gateway deployment.

PR #776, opened 2025-07-16
fix(tap-agent): check horizon is active before checking for escrow

suchapalaver:
Problem: The tap-agent was crashing because it tried to create the escrow_accounts_v2 channel before checking if Horizon (V2) was active. Since V2 is not active, the network subgraph doesn't have the paymentsEscrowAccounts field, causing the GraphQL query to fail.

Thanks @rotarur fr reporting!

### DIFF
--- a/crates/tap-agent/src/agent.rs
+++ b/crates/tap-agent/src/agent.rs
@@ -40,7 +40,8 @@ use indexer_config::{
     SubgraphConfig, SubgraphsConfig, TapConfig,
 };
 use indexer_monitor::{
-    escrow_accounts_v1, escrow_accounts_v2, indexer_allocations, DeploymentDetails, SubgraphClient,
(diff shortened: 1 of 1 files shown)


neithanmo (approved):

crates/tap-agent/src/agent.rs
```diff
@@ -205,13 +196,29 @@ pub async fn start_agent() -> (ActorRef<SenderAccountsManagerMessage>, JoinHandl
...
+    // V2 escrow accounts are in the network subgraph, not a separate TAP v2 subgraph
+    let escrow_accounts_v2 = if is_horizon_enabled {
+        escrow_accounts_v2(
+            network_subgraph,
+            *indexer_address,
+            *network_sync_interval,
+            false,
+        )
+        .await
+        .expect("Error creating escrow_accounts_v2 channel")
```
ryan-s-roberts, line 209: This is a reasonable crash as it's a feature mismatch?

ryan-s-roberts (approved):

PR #787, opened 2025-07-21
test: replace sqlx test with testcontainers

suchapalaver:


crates/tap-agent/src/agent/sender_account.rs
```diff
@@ -2013,7 +2013,6 @@ pub mod tests {
     }
 
     #[tokio::test]
-    #[ignore = "Retry mechanism test is flaky - needs investigation of timing/mock setup"]
```
suchapalaver: These will pass - or at least, they pass on my machine! - when run on their own.

crates/tap-agent/src/agent/sender_account.rs
```diff
@@ -2444,7 +2443,6 @@ pub mod tests {
     }
 
     #[tokio::test]
-    #[ignore = "Escrow account state management test is flaky after ractor migration"]
```
suchapalaver: These will pass - or at least, they pass on my machine! - when run on their own.

crates/test-assets/src/lib.rs
```diff
@@ -501,3 +502,178 @@ pub fn pgpool() -> Pin<Box<dyn Future<Output = PgPool>>> {
...
+    pub url: String,
+    _container: ContainerAsync<postgres::Postgres>,
+}
+
+/// Set up a test database using testcontainers
+///
+/// This creates an isolated PostgreSQL container and database for testing.
+/// The container will be kept alive as long as the returned TestDatabase
+/// instance is not dropped.
+pub async fn setup_shared_test_db() -> TestDatabase {
```
neithanmo, line 525: this is magic!
I will add something like this to graph-agent for sure!

PR #912, opened 2026-01-27
feat(monitor): add number_gte support with tiered backoff to allocation queries

MoonBoi9001:
## Summary

- Adds `number_gte` parameter to allocation queries to ensure responses come from indexers synced to at least the last successful block - meaning the block height of the last respond we got from the gateway.
- Implements tiered backoff strategy when Gateway returns "block too high" errors (when indexers have not synced to at least the block height of the previous response)
- This RP effectively protects against overwriting valid allocation data with empty results from stale indexers that are serving old data.

## Problem

The Graph Gateway routes queries to different indexers with varying sync states. Without block freshness constraints, a query might be served by a stale indexer that returns 0 allocations (because it hasn't indexed them yet), overwriting valid cached data.

## Solution

Track the block number from successful responses and use `number_gte` to constrain subsequent queries. When queries fail due to indexers being behind, progressively relax the constraint through tiers:

| Tier | Relax the block height constraint by | Equilavalent time on Arbitrum |
|------|---------|---------------|
| 0 | 0 blocks | immediate |
| 1 | 100 blocks | ~25 sec |
| 2 | 1,000 blocks | ~4 min |
| 3 | 10,000 blocks | ~40 min |
| 4+ | Just use block height from error message | last resort |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4289,6 +4289,7 @@ dependencies = [
  "indexer-attestation",
  "indexer-query",
  "indexer-watcher",
+ "regex",
  "reqwest 0.12.24",
  "rstest",
  "serde",
--- a/crates/monitor/Cargo.toml
+++ b/crates/monitor/Cargo.toml
@@ -19,6 +19,7 @@ serde = { workspace = true, features = ["derive"] }
 serde_json.workspace = true
 tokio.workspace = true
 bip39.workspace = true
+regex.workspace = true
 
 [dev-dependencies]
 env_logger.workspace = true
--- a/crates/monitor/src/allocations.rs
+++ b/crates/monitor/src/allocations.rs
@@ -23,6 +23,11 @@ use crate::{
 /// Receiver of Map between allocation id and allocation struct
 pub type AllocationWatcher = Receiver<HashMap<Address, Allocation>>;
 
+/// Minimum expected block number for Arbitrum.
+/// Used as a sanity check to detect stale indexers on initial query.
+/// If the initial response returns a block below this, we retry.
+const ARBITRUM_MIN_EXPECTED_BLOCK: i64 = 400_000_000;
+
 /// An always up-to-date list of an indexer's active and recently closed allocations.
 ///
 /// Uses tiered backoff to handle Gateway routing to indexers with varying sync states.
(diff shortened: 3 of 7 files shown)


neithanmo (changes requested):
While I understand the intent to handle Gateway routing to out-of-sync indexers, this PR has issues that block merging:

1. Hardcoded `ARBITRUM_MIN_EXPECTED_BLOCK = 400_000_000` - Breaks on Base, Sepolia, local dev, and any non-Arbitrum-mainnet chain.
2. Brittle error parsing - String matching on "missing block" / "latest: \d+" is not a stable contract. Gateway error formats can change without notice.
3. The check "empty = stale" logic - The `should_update` check rejects empty results even from fresh indexers? this needs to be double check tho. If an indexer legitimately closes all allocations, the watcher will preserve stale data indefinitely.

**On the broader approach**: The current behavior without this PR is recoverable - queries fail, get logged, and retry until Gateway routes to a synced indexer.


We do think this issue must be addressed on the gateway side, because this problem is not severe enough to warrant a client-side workaround.

@MoonBoi9001 We are investigating the Gateway side to figure out what is  the best approach. 

Thanks for digging into this and putting together a detailed solution - the tiered backoff concept is solid thinking. Appreciate the effort to improve reliability here! 🙏

MoonBoi9001:
>The current behavior without this PR is recoverable - queries fail, get logged, and retry until Gateway routes to a synced indexer.

True, however there's an issue here that indexers aren't able to serve queries for their subgraphs when their allocation data is served back to them incorrectly, since their indexer-rs beleives that any receipts it receives from the gateway are not correct, so it will not accept them and respond to incomming queries. Indexers that run their own version of the network subgraph can bypass this, but for those that don't it appears to us when we look inside bigquery that they have a really low success rate. Also this affects their score as determined by both the ISA for query time selection and IISA for DIPS, the ISA will start to deprioritise them affecting how much revenue they can generate from serving queries, since from it's perspective it sees an indexer that cannot respond to queries it sends. 


>It will be wrong for other chains like Base, Base sepolia, local testnets we use for testing and so on.

Okay I was not aware that the protocol also lives on base & Base sepolia and those chains are being used for testing, makes sense that ARBITRUM_MIN_EXPECTED_BLOCK should not be hard coded, but also forgot that this would have issues with arb sepolia & local dev env.



>Brittle error parsing - String matching on "missing block" / "latest: \d+" is not a stable contract. Gateway error formats can change without notice.
The check "empty = stale" logic - The should_update check rejects empty results even from fresh indexers? this needs to be double check tho. If an indexer legitimately closes all allocations, the watcher will preserve stale data indefinitely.

Appreciate the review & other comments here, I would be happy to address all review feedback and put this PR in draft for now until that's done, but do you think from your earlier comment "this issue must be addressed on the gateway side" that this repo is not the right place for such a PR then?

PR #919, opened 2026-02-02
Fix TapReceipt type mismatch in allocation middleware

neithanmo:
Fixes a bug where the allocation middleware was unable to extract the allocation ID from TAP receipts due to a TypeId mismatch, causing incorrect allocation selection and attestation failures.

## Issue

The allocation middleware was looking for `SignedReceipt` in request extensions, but `receipt_middleware` inserts the `TapReceipt` enum wrapper. This `TypeId` mismatch caused `extensions.get::<SignedReceipt>()` to always return None, forcing a fallback to the `deployment_to_allocation` map.

When multiple allocations exist for the same deployment, this could cause:
•  Attestations signed with the wrong allocation's derived key
•  Gateway verification failures ("bad attestation" errors)
•  Incorrect metrics attribution

The Solution is simple:

•  Update the middleware to look for TapReceipt instead of SignedReceipt
•  Handle both V1 and V2 receipt variants:
◦  V1: Extract allocation_id directly from the receipt message
◦  V2: Convert collection_id to address via CollectionId::as_address()

### Impact

This issue might cause signature verification errors when validating indexer's attestation, due to the gateway using a different allocation than the indexer. This happens only if there are more than one allocation/collection per deployment, otherwise, this issue wont cause any issues and would pass silently.

### DIFF
--- a/crates/service/src/middleware/allocation.rs
+++ b/crates/service/src/middleware/allocation.rs
@@ -9,9 +9,9 @@ use axum::{
     response::Response,
 };
 use thegraph_core::{alloy::primitives::Address, CollectionId, DeploymentId};
+use tokio::sync::watch;
 
 use crate::tap::TapReceipt;
-use tokio::sync::watch;
 
 /// The current query Allocation ID address
 // TODO: Use thegraph-core::AllocationId instead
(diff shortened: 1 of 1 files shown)


suchapalaver (changes requested):
Great catch @neithanmo 🙌 . Just need the formatting and clippy fixed before we can approve 🫡

suchapalaver (approved):

PR #968, opened 2026-03-04
feat: log proposal rejection reasons on the indexer side

MoonBoi9001:
## TL;DR

When the indexer rejects a paid indexing proposal, today nothing about why is recorded on the indexer side — only the consumer sees the reason. This change writes the rejection reason and proposed deployment ID to the indexer's logs at info level. The change is logging-only, with no behaviour or wire format changes.

## Motivation

When a payer submits a paid indexing agreement proposal to the indexer, the indexer's service responds over its proposal API with a reject reason but writes nothing to its own logs. An operator looking at the indexer's logs alone can see that proposals are flowing in and being rejected, but cannot tell which subgraph deployment was being proposed or why each one was turned down. Diagnosing a wave of rejections then requires correlating with the consumer side, which an indexer operator may not have access to.

## Summary

- Add a helper that does best-effort decoding of a proposal to recover the deployment ID
- Record the reject reason, deployment ID, and error text on the rejection log line
- Drop the rejection log level from warn to info, since rejections are normal protocol flow

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -350,6 +350,17 @@ fn bytes32_to_ipfs_hash(bytes: &[u8; 32]) -> String {
     bs58::encode(&multihash).into_string()
 }
 
+/// Try to extract the deployment ID from raw signed RCA bytes.
+///
+/// Best-effort: returns `None` if any decoding step fails.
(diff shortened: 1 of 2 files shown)


MoonBoi9001:

PR #983, opened 2026-03-18
fix: accept v1 indexing agreement proposals during validation

MoonBoi9001:
## TL;DR

The indexer rejects every legitimate v1 paid indexing proposal because of a mismatch with how the on-chain contract encodes the agreement version. The fix changes the validator to match the contract's encoding. This unblocks the end-to-end paid indexing flow.

## Motivation

When the indexer receives a paid indexing proposal, its proposal validator decodes the agreement metadata and only continues if the metadata's version field equals one. The on-chain contract encodes v1 of the agreement as zero, not one — the smart-contract enum lists v1 as its first variant, which the ABI represents as zero. So every legitimate v1 proposal is rejected as "unsupported metadata version", and the paid indexing runtime flow is fully blocked at the first validation step. This change pairs with the consumer-side fix in edgeandnode/dipper#583.

## Summary

- Match the validator on the contract's encoded value instead of the wrong literal
- Update the test fixtures that asserted the wrong value
- Pairs with the consumer-side fix in edgeandnode/dipper#583

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -410,8 +410,8 @@ pub async fn validate_and_create_rca(
                 ))
             })?;
 
-    // Only support version 1 terms for now
-    if metadata.version != 1 {
+    // Only support V1 terms (IndexingAgreementVersion.V1 = 0 in Solidity enum)
+    if metadata.version != 0 {
         return Err(DipsError::UnsupportedMetadataVersion(metadata.version));
     }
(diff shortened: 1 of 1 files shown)


RembrandtK (approved):
Would be nice to have an enum or constants instead of magic numbers. :) 

I think I noticed this bug too, but did not have time to verify or create PR. Definitely should help. :)

PR #988, opened 2026-03-26
feat: log indexer pricing configuration at startup

MoonBoi9001:
## TL;DR

When the paid indexing component starts, the indexer doesn't log what pricing configuration it loaded. This change writes the supported networks, the per-network minimum prices, and the entity pricing to the log at startup. The change is logging-only.

## Motivation

When a paid-indexing-enabled indexer starts, it logs a warning only if its supported-networks list is empty. With a populated list — the normal case — nothing is written about which networks the indexer is willing to be paid on, what minimum price it expects on each, or what entity pricing applies. An operator deploying to production cannot verify from the logs that the pricing configuration was parsed correctly, and the only way to confirm today is to hit a status HTTP endpoint after startup.

## Summary

- Log the configured supported networks at info level on startup
- Log the contract address and IPFS endpoint the component will use
- Log the per-network minimum price, in GRT, alongside each supported network
- Log the entity-pricing configuration when present

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -350,6 +350,17 @@ fn bytes32_to_ipfs_hash(bytes: &[u8; 32]) -> String {
     bs58::encode(&multihash).into_string()
 }
 
+/// Try to extract the deployment ID from raw signed RCA bytes.
+///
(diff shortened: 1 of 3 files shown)


MoonBoi9001:

PR #990, opened 2026-03-26
feat: include chain identity in price rejection logs

MoonBoi9001:
## TL;DR

Price rejection logs today carry the network name but not the standard chain identifier (the CAIP-2 form like eip155:42161 for Arbitrum One). This change adds the chain identifier and converts the offered and minimum prices from text-string interpolation to structured log fields. Operators can now filter rejections by chain and by price values directly.

## Motivation

When the indexer rejects a paid indexing proposal because the offered price is below the configured minimum, it writes a log line that names the network and embeds the offered and minimum prices inside a free-form text message. An operator running multiple chains cannot filter rejection logs by chain without parsing the message text, and the same applies to filtering by price values. This makes triaging rejections on a multi-chain indexer slower than it needs to be.

## Summary

- Resolve the standard chain identifier from the networks registry or extra-networks config
- Include the chain identifier on every price-related rejection log
- Promote offered and minimum prices to structured fields so they are queryable directly
- Apply the same change to both the per-second and per-entity-per-second rejection paths

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -630,6 +630,55 @@ mod test {
         assert_eq!(id.as_bytes(), &expected_hash[..16]);
     }
 
+    /// Shared test vector with dipper (dipper-rpc/src/indexer.rs).
+    /// Both repos must produce the same bytes16 for this input.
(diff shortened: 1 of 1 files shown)


MoonBoi9001:

PR #1009, opened 2026-04-15
feat: switch to on-chain offer-based authorization

MoonBoi9001:
## TL;DR

Authorization for paid indexing agreements moves from an off-chain signature check inside the indexer to an on-chain offer registered by the payer, which the smart contract verifies at acceptance time. This is the indexer-side half of that switch; the consumer half ships as dipper #607. Indexers without escrow accounts will no longer be auto-rejected during the proposal step.

## Motivation

When a payer proposes a paid indexing agreement to an indexer, the indexer's service today recovers the signature on the proposal and rejects anything not signed by an address registered in on-chain escrow. The check duplicates what the smart contract already enforces at acceptance time, and it has a side effect: indexers without any escrow accounts get rejected during negotiation before they can even respond on terms.

A new contracts branch shifts authorization to an on-chain step where the payer registers an offer before the agreement can be accepted, and the contract checks the offer's hash at acceptance. That fits the real order of operations — proposals come first, the payer (in dipper #607) only registers the offer after the indexer has accepted terms — so the indexer-side signature check becomes redundant and is removed.

## Summary

- Remove the indexer-side signature check from proposal validation; the contract enforces it later
- Delete the now-unused signer-validation logic and its tests
- Drop the escrow-watcher dependency from the proposal handler since the check it fed has gone
- Add a conditions placeholder to the agreement layout so it matches the audited contract
- Keep the existing rejection reason on the wire for compatibility, even though no code emits it
- Rewrite the handler tests on a small wire-bytes helper; add a regression test for the new layout


crates/config/maximal-config-example.toml
```diff
@@ -174,17 +174,18 @@ max_receipts_per_request = 10000
 0xDDE4cfFd3D9052A9cb618fC05a1Cd02be1f2F467 = "https://tap-aggregator.network.thegraph.com"
 0xDD6a6f76eb36B873C1C184e8b9b9e762FE216490 = "https://tap-aggregator-arbitrum-one.graphops.xyz"
 
-# DIPS (Decentralized Indexing Payment System)
-# NOTE: DIPS requires Horizon mode ([horizon].enabled = true)
-# Payer authorization is handled via escrow accounts (same trust model as TAP)
-#
-# Pricing uses human-readable GRT values (not wei), e.g. "100" = 100 GRT per 30 days.
+# DIPS (Decentralized Indexing Payment System). Requires Horizon mode.
```
RembrandtK, line 177: Claude hallucination for DIPS? :) In other places too.

Is there a Horizon mode BTW, or always operating that way?

RembrandtK (approved):

crates/config/maximal-config-example.toml
MoonBoi9001: Always operating that way now I believe. Will update this comment

Issue #1013, opened 2026-04-17
Config changes


tmigone:
**Remove `horizon.enabled`**
Its now mandatory to be true and the service crashes if it is not (need to verify this) so just remove it and asume default behavior.

**Remove `graph_tally_collector_address`**
```
[blockchain]
chain_id = 1337
receipts_verifier_address = "${tap_verifier}"
receipts_verifier_address_v2 = "${graph_tally_verifier}"
subgraph_service_address = "${subgraph_service}"
graph_tally_collector_address = "${graph_tally_verifier}"
```

`receipts_verifier_address_v2` and `graph_tally_collector_address` are the same thing. We should remove `graph_tally_collector_address` because its the latest addition, whereas `receipts_verifier_address_v2` was added with horizon and indexers have it configured already.

PR #1014, opened 2026-04-17
fix(config): several configuration optimizations

tmigone:
Addresses #1013 

Horizon mode is now always enabled, making several config fields and abstractions unnecessary. This
PR removes the dead code while maintaining backwards compatibility for existing configs.

## Changes

Remove `graph_tally_collector_address`
- Duplicate of `receipts_verifier_address_v2`; use the latter instead

Deprecate `horizon.enabled`
- Field is now ignored if present (no breaking change for existing configs)
- Removed validation that rejected `enabled = false`

Remove `TapMode` enum
- Only had one variant (Horizon), so replaced with direct `Address` usage
- `SenderAccountConfig.tap_mode` → `SenderAccountConfig.subgraph_service_address`
- Removed `Config.tap_mode()` method

Simplify `subgraph_service_address`
- `BlockchainConfig.subgraph_service_address: Option<Address> → Address`
- `TapAgentContext.subgraph_service_address: Option<Address> → Address`
- Removed unnecessary getter methods that just returned the field

### DIFF
--- a/crates/config/maximal-config-example.toml
+++ b/crates/config/maximal-config-example.toml
@@ -193,7 +193,6 @@ max_receipts_per_request = 10000
 0xDD6a6f76eb36B873C1C184e8b9b9e762FE216490 = "https://tap-aggregator-arbitrum-one.graphops.xyz"
 
 # DIPS (Decentralized Indexing Payment System)
-# NOTE: DIPS requires Horizon mode ([horizon].enabled = true)
 [dips]
 host = "0.0.0.0"
 port = "7601"
--- a/crates/config/maximal-config-example.toml
+++ b/crates/config/maximal-config-example.toml
@@ -110,7 +110,7 @@ syncing_interval_secs = 60
 [blockchain]
 # The chain ID of the network that the graph network is running on
 chain_id = 1337
-# Horizon addresses; required when [horizon].enabled = true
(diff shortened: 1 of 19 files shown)


Maikol (dismissed):

crates/config/maximal-config-example.toml
MoonBoi9001: could update this since horizon always enabled

crates/config/src/config.rs
```diff
@@ -831,6 +761,40 @@ mod tests {
         assert_eq!(max_config, max_config_file);
     }
 
+    /// Test backwards compatibility: configs with deprecated [horizon] section should still work
+    #[sealed_test(files = ["minimal-config-example.toml", "default_values.toml"])]
+    #[allow(deprecated)]
+    fn test_horizon_enabled_backwards_compatibility() {
```
MoonBoi9001, line 767: suggestion: could parse the indexer TOML instead of using env var

MoonBoi9001 (dismissed):

crates/config/src/config.rs
tmigone: true, i think for the test it doesnt matter too much tho

MoonBoi9001 (approved):

PR #1015, opened 2026-04-17
feat: remove v1 escrow

tmigone:


docs/migration-config/arbitrum-sepolia-config.toml
```diff
@@ -19,12 +19,6 @@ status_url = "${INDEXER_SERVICE_GRAPH_NODE_STATUS_ENDPOINT}"
...
 deployment_id = "${INDEXER_SERVICE_NETWORK_SUBGRAPH_DEPLOYMENT}"
 
-# [subgraphs.escrow]
-# INDEXER_SUBGRAPHS__ESCROW__QUERY_URL
-# query_url = ""
-# INDEXER_SUBGRAPHS__ESCROW__DEPLOYMENT_ID
-# deployment_id = ""
-
 [blockchain]
 chain_id = 421614
```
MoonBoi9001, line 23: I think we should add `receipts_verifier_address_v2` and `subgraph_service_address`

crates/config/src/config.rs
```diff
@@ -306,11 +296,46 @@ impl Config {
...
+            tracing::warn!(
+                "The `subgraphs.escrow` configuration is deprecated and will be ignored. \
+                V2 escrow accounts are now sourced from the network subgraph. \
+                You can safely remove the `[subgraphs.escrow]` section from your config."
+            );
+        }
+
+        // Warn about deprecated serve_escrow_subgraph
+        #[allow(deprecated)]
+        if self.service.serve_escrow_subgraph {
```
MoonBoi9001: We should change `serve_escrow_subgraph: bool` to `serve_escrow_subgraph: Option<bool>` in line 585 otherwise this wont fire with `serve_escrow_subgraph = false`

contrib/tap-agent/start.sh
MoonBoi9001, line 69: this fetches the legacy V1 escrow subgraph ID but the config no longer uses it

MoonBoi9001 (dismissed):

contrib/tap-agent/start.sh
tmigone: `contrib/` should honestly be removed. This was used by semiotic to run local-network in some capacity. I dont think we need it but I want to keep it for a bit because it has some performance testing set up that might be useful.

MoonBoi9001 (approved):

PR #1022, opened 2026-04-23
perf: explicitly cast char(n) types in tap receipts table queries

tmigone:
Add explicit `::char(n)` casts to SQL query parameters to match the column types in `tap_horizon_receipts` and `tap_horizon_ravs` tables. This prevents PostgreSQL from performing type coercion when comparing text parameters to `char(n)` columns, allowing proper index usage and reducing query time substantially (observed ~30s to <1ms).

### DIFF
--- a/crates/tap-agent/src/agent/sender_allocation.rs
+++ b/crates/tap-agent/src/agent/sender_allocation.rs
@@ -967,11 +967,11 @@ impl DatabaseInteractions for SenderAllocationState<Horizon> {
             FROM
                 tap_horizon_receipts_invalid
             WHERE
-                collection_id = $1
-                AND service_provider = $2
-                AND payer = $3
-                AND data_service = $4
-                AND signer_address IN (SELECT unnest($5::text[]))
+                collection_id = $1::char(64)
+                AND service_provider = $2::char(40)
+                AND payer = $3::char(40)
+                AND data_service = $4::char(40)
+                AND signer_address IN (SELECT unnest($5::char(40)[]))
             "#,
         )
         // self.allocation_id is already a CollectionId in Horizon state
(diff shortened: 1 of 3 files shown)


crates/tap-agent/src/agent/sender_allocation.rs
```diff
@@ -967,11 +967,11 @@ impl DatabaseInteractions for SenderAllocationState<Horizon> {
...
-                collection_id = $1
-                AND service_provider = $2
-                AND payer = $3
-                AND data_service = $4
-                AND signer_address IN (SELECT unnest($5::text[]))
+                collection_id = $1::char(64)
+                AND service_provider = $2::char(40)
+                AND payer = $3::char(40)
+                AND data_service = $4::char(40)
+                AND signer_address IN (SELECT unnest($5::char(40)[]))
```
lutter: This should be `signer_address = any($5::char(40)[])`

crates/tap-agent/src/agent/sender_allocation.rs
```diff
@@ -1026,12 +1026,12 @@ impl DatabaseInteractions for SenderAllocationState<Horizon> {
...
-                AND service_provider = $2
-                AND payer = $3
-                AND data_service = $4
+                collection_id = $1::char(64)
+                AND service_provider = $2::char(40)
+                AND payer = $3::char(40)
+                AND data_service = $4::char(40)
                 AND id <= $5
-                AND signer_address IN (SELECT unnest($6::text[]))
+                AND signer_address IN (SELECT unnest($6::char(40)[]))
```
lutter: same comment - use `= any`

lutter (approved):

PR #1037, opened 2026-05-20
feat: advertise DIPs prices via /dips/info endpoint

MoonBoi9001:
## TL;DR

A new public HTTP endpoint surfaces an indexer's DIPs pricing and supported chains as JSON, populated from three opt-in config fields. The endpoint is signalling-only, so the existing payment validation path stays untouched. Indexers without the new fields configured see no behavioural change.

## Motivation

Before DIPs can roll out to any chains, we need to know which chains indexers are willing to support and at what minimum price. Today there is no machine-readable way for an indexer to advertise this, so we cannot scan the indexer set to see where coverage is thick enough for a rollout to be viable.

This change gives every indexer running the service a way to publish their supported networks and minimum prices in configuration, exposed at a public HTTP endpoint. External tooling can then read across the indexer set to identify which chains have enough support to be good rollout candidates.

## Summary

- Add three opt-in `[dips]` fields: supported networks, monthly minimum per network, per entity.
- Expose them at `GET /dips/info` as JSON whenever `[dips]` is configured.
- Leave the existing payment validation path untouched; new fields feed only the new endpoint.
- Drop legacy pricing fields from example; struct keeps them defaulted so old configs still parse.
- Ship a commented-out pricing table for 90+ networks in the example config.

Co-Authored-By: Claude Opus 4.7 (1M context) <noreply@anthropic.com>


crates/config/src/config.rs
```diff
@@ -669,11 +669,33 @@ fn default_allocation_reconciliation_interval_secs() -> Duration {
 pub struct DipsConfig {
```
tmigone, line 663: Doesnt have to be addressed on this PR, but can we add documentation for each one of DipsConfig elements?

crates/config/src/grt.rs
```diff
@@ -4,6 +4,53 @@
...
+#[derive(Debug, PartialEq, Default, Clone, Copy)]
+pub struct GRT(u128);
+
+impl GRT {
+    pub const ZERO: GRT = GRT(0);
+
+    /// Convert GRT string to wei for test construction.
+    /// Panics on invalid input - only use in tests.
+    #[cfg(test)]
+    pub fn from_grt(grt: &str) -> Self {
```
tmigone: Why is this function needed? I don't see it being used in the project.

crates/config/maximal-config-example.toml
```diff
@@ -175,16 +175,123 @@ max_receipts_per_request = 10000
...
+#
+# For reference: analysis of subgraphs indexed by the upgrade indexer in Q1 2025 found
+# the average entity size to be ~0.759 KiB. At this size, 1 billion entities ≈ 0.707 TiB.
+# Your own observations may differ - adjust pricing accordingly.
+# min_grt_per_billion_entities_per_30_days = "200"  # entity-based component (global)
 
-[dips.additional_networks]
-"eip155:1337" = "hardhat"
+[dips.min_grt_per_30_days]  # base rate component (per-network)
+# mainnet = "45"
```
tmigone: I think just listing a couple examples here would be enough. Since the network list is dynamic no need to make it comprehensive here.

crates/config/maximal-config-example.toml
```diff
@@ -175,16 +175,123 @@ max_receipts_per_request = 10000
...
 [dips]
 host = "0.0.0.0"
 port = "7601"
-allowed_payers = ["0x3333333333333333333333333333333333333333"]
 
-price_per_entity = "1000"
+# Networks you explicitly support indexing.
+# See https://github.com/graphprotocol/networks-registry/blob/main/docs/networks-table.md
+# e.g. supported_networks = ["mainnet", "arbitrum-one"]
+supported_networks = []
```
tmigone: It's not clear what the default behavior is if a network is included in `supported_networks` but not given a price via `dips.min_grt_per_30_days`. Almost feels like both should be merged maybe?

crates/service/src/service.rs
```diff
@@ -132,6 +152,25 @@ pub async fn run() -> anyhow::Result<()> {
...
+        // intent — render it verbatim rather than deriving from priced chains, so
+        // declaring a chain without pricing still surfaces in the published signal.
+        let mut supported_networks: Vec<String> = dips.supported_networks.iter().cloned().collect();
+        supported_networks.sort();
+        DipsInfoState {
+            supported_networks,
+            min_grt_per_30_days: dips
+                .min_grt_per_30_days
+                .iter()
+                .map(|(network, grt)| (network.clone(), format_grt(grt.wei())))
```
tmigone: why do we need human readable formating for grt amounts here? The endpoint is meant to be consumed by some other component primarily no?

crates/service/src/service.rs
```diff
@@ -132,6 +152,25 @@ pub async fn run() -> anyhow::Result<()> {
         }
     };
 
+    let dips_info_state = config.dips.as_ref().map(|dips| {
+        // Sort for deterministic JSON output. `supported_networks` is the indexer's
```
tmigone: why is it important to have deterministic output here? also not sorting will also be deterministic just un-sorted 😄

crates/service/src/routes/dips_info.rs
tmigone, line 1: Don't know what can be simplified here but this entire file seems a bit redundant. 

`DipsInfoState` is a subset of `DipsConfig`, then `DipsInfoPricing` is a subset of `DipsInfoState`, and then `DipsInfoResponse` is just another way of re-arranging the same information 😄

crates/service/src/service.rs
MoonBoi9001: This is fair, but can also be useful for analytics of competitors if you're an indexer and thinking about how to price your services, this makes the endpoint slightly more user readable and consistent with what indexers actually configured in their config, but I could go either way with this.

crates/config/src/config.rs
MoonBoi9001: Addressed in fbb36ce

crates/config/maximal-config-example.toml
MoonBoi9001: Addressed in c7647c9
A few more than a couple, but I think it gets the ball rolling on analytics

crates/config/src/grt.rs
MoonBoi9001: It came in when I copied the file across from `main-dips`. A test was using it briefly, but when I later emptied out the example config and simplified the test to match, nothing was using it any more. Deleted in d79bf85.

crates/service/src/routes/dips_info.rs
MoonBoi9001: Addressed in 2349636

crates/config/maximal-config-example.toml
MoonBoi9001: Dropped `supported_networks` entirely in 67debf7

crates/config/maximal-config-example.toml
```diff
@@ -175,16 +175,33 @@ max_receipts_per_request = 10000
 0xDD6a6f76eb36B873C1C184e8b9b9e762FE216490 = "https://tap-aggregator-arbitrum-one.graphops.xyz"
 
 # DIPS (Decentralized Indexing Payment System)
+# These fields are consumed by the public `/dips/info` HTTP endpoint. External
+# tooling reads them to discover which networks you are willing to index and at
+# what minimum price, ahead of any indexing agreement actually being offered.
+#
+# Pricing uses human-readable GRT values (not wei), e.g. "100" = 100 GRT per 30 days.
+# Pricing a chain in [dips.min_grt_per_30_days] is also your public declaration
+# of support for that chain — there is no separate supported_networks list.
```
tmigone:
```suggestion
# of support for that chain.
```

crates/config/maximal-config-example.toml
```diff
@@ -175,16 +175,33 @@ max_receipts_per_request = 10000
...
-
-price_per_entity = "1000"
 
-[dips.price_per_epoch]
-mainnet = "100"
-hardhat = "100"
-
-[dips.additional_networks]
-"eip155:1337" = "hardhat"
+# Minimum payment you are willing to accept in order to accept indexing agreements
```
tmigone: this is still a bit confusing. its not clear how `min_grt_per_billion_entities_per_30_days` and `min_grt_per_30_days` relate to the formula expressed here

crates/service/src/routes/dips_info.rs
```diff
@@ -0,0 +1,90 @@
+// Copyright 2023-, Edge & Node, GraphOps, and Semiotic Labs.
+// SPDX-License-Identifier: Apache-2.0
+
+use axum::{extract::State, Json};
+use serde::Serialize;
+use std::collections::BTreeMap;
+
+/// Indexer DIPs pricing signal. Serves both as the `/dips/info` JSON response
+/// and as the axum State held by the handler — derived once from `DipsConfig`
+/// at startup and served as-is on every request.
```
tmigone:
```suggestion
/// Indexer DIPs pricing signal. Derived from `DipsConfig`
/// at startup and served as-is on every request.
```

crates/service/src/service/router.rs
```diff
@@ -84,6 +85,10 @@ pub struct ServiceRouter {
     network_subgraph: Option<(&'static SubgraphClient, NetworkSubgraphConfig)>,
     allocations: Option<AllocationWatcher>,
     dispute_manager: Option<DisputeManagerWatcher>,
+    // Optional state for the public `/dips/info` endpoint. Populated only when
+    // `[dips]` is set in config; consumed purely for signalling indexer pricing
+    // and supported networks, never feeds the gRPC proposal handler.
```
tmigone, line 90: > never feeds the gRPC proposal handler.

what does this mean? Isn't it the opposite?

crates/service/src/service.rs
```diff
@@ -132,6 +152,24 @@ pub async fn run() -> anyhow::Result<()> {
         }
     };
 
+    let dips_info_state = config.dips.as_ref().map(|dips| {
+        let min_grt_per_30_days: BTreeMap<String, String> = dips
+            .min_grt_per_30_days
+            .iter()
+            .map(|(network, grt)| (network.clone(), format_grt(grt.wei())))
```
tmigone, line 159: Do we need formatting values to be human readable? If this is meant to be consumed by the dipper primarily I would skip this and just use the raw wei value, presumably the dipper is parsing back on the other end anyways.

crates/service/src/service/router.rs
tmigone: ah the gRPC handler is a different service I guess legacy?

crates/service/src/service.rs
MoonBoi9001: Thinking that human readable is more useful for analytics of competitors if you're an indexer and thinking about how to price your services against competition.

tmigone: just saw you already replied this 2 weeks ago 😆 sounds good then!

crates/config/maximal-config-example.toml
MoonBoi9001: I'll ammend that comment to make it more clear

crates/service/src/service/router.rs
MoonBoi9001: > never feeds the gRPC proposal handler.

> what does this mean? Isn't it the opposite?

In the state of the code introduced by this pr then, nothing is actually wired up to the gRPC service that dipper will be sending agreements to. This stuff, at this point in time, is analytics only. So the comment reflects todays reality after PR merges.

tmigone (approved):
LGTM

PR #1041, opened 2026-06-05
feat: separate accept and reject in the DIPs proposal response

MoonBoi9001:
## TL;DR

When the indexer answers a DIPs proposal it returns one of two outcomes: accepted or rejected. Until now both shared one response shape, so an accept had to carry a meaningless "no reason" reject code. This reshapes the response so accept and reject are distinct, and a reject carries a reason plus optional detail.

## Motivation

The indexer's response to a DIPs proposal packed the accept/reject answer and the rejection reason into one flat message. An accept therefore had to set the reason field to a placeholder default, which read as if a reject reason applied to an accept — confusing for anyone reading the contract, and a field that could only ever be ignored.

Separating the two outcomes removes that dead placeholder: an accept now carries nothing it doesn't need, and the reject reason only exists where it's meaningful. That also lets the catch-all reason double as "rejected for a reason not yet named," so the sender (the Dipper service) can still act on rejections it has no specific code for.

## Summary

- Model the response as one of two outcomes: an accepted or a rejected variant.
- A rejected outcome carries a reason code plus an optional free-text detail.
- An accepted outcome carries no reason field, so accept and reject no longer share one.
- Fold the old generic reason into the catch-all default (rejected, reason not listed).
- Map store/DB failures to a transient reason so the sender knows it can safely resend.
- Sanitise internal store errors in the detail so they don't leak to the sender.
- Regenerate the committed gRPC bindings from the updated protocol file.

### DIFF
--- a/crates/dips/proto/indexer.proto
+++ b/crates/dips/proto/indexer.proto
@@ -24,34 +24,51 @@ message SubmitAgreementProposalRequest {
 /**
  * A response to a request to propose a new _indexing agreement_ to an _indexer_.
  *
+ * The outcome is either `accepted` or `rejected` -- exactly one is set. A
+ * rejection carries a `RejectReason` plus an optional free-text detail.
+ *
  * See the `DipsService.SubmitAgreementProposal` method.
  */
 message SubmitAgreementProposalResponse {
-  ProposalResponse response = 1; /// The response to the agreement proposal.
-  RejectReason reject_reason = 2; /// Only set when response = REJECT.
+  oneof outcome {
+    Accepted accepted = 1; /// Set when the proposal was accepted.
+    Rejected rejected = 2; /// Set when the proposal was rejected.
+  }
 }
 
 /**
- * The response to an _indexing agreement_ proposal.
+ * An accepted _indexing agreement_ proposal. Empty for now; accept-only fields
+ * can be added later without disturbing the reject path.
  */
-enum ProposalResponse {
-  ACCEPT = 0; /// The agreement proposal was accepted.
-  REJECT = 1; /// The agreement proposal was rejected.
+message Accepted {}
+
+/**
+ * A rejected _indexing agreement_ proposal.
+ */
+message Rejected {
+  RejectReason reason = 1; /// Why the proposal was rejected.
+  string detail = 2;       /// Optional human-readable detail; may be empty.
 }
 
 /**
- * The reason for rejecting an _indexing agreement_ proposal.
- * Only meaningful when ProposalResponse = REJECT.
+ * The reason an _indexer_ rejected an _indexing agreement_ proposal. Values may
+ * be added over time; an older reader should treat an unrecognised reason as
+ * UNSPECIFIED (the catch-all).
  */
 enum RejectReason {
-  REJECT_REASON_UNSPECIFIED = 0;              /// Default / not set (used for ACCEPT responses).
-  REJECT_REASON_PRICE_TOO_LOW = 1;            /// The offered price is below the indexer's minimum.
-  REJECT_REASON_OTHER = 2;                    /// Any other reason (bad signature, etc.).
-  REJECT_REASON_SIGNER_NOT_AUTHORISED = 3;    /// The proposal signer is not authorised on the escrow contract.
-  REJECT_REASON_DEADLINE_EXPIRED = 4;         /// The proposal deadline has already passed.
-  REJECT_REASON_UNSUPPORTED_NETWORK = 5;      /// The subgraph's network is not supported by this indexer.
-  REJECT_REASON_SUBGRAPH_MANIFEST_UNAVAILABLE = 6; /// The subgraph manifest could not be fetched from IPFS.
-  REJECT_REASON_UNEXPECTED_SERVICE_PROVIDER = 7;   /// The RCA service provider does not match this indexer.
-  REJECT_REASON_AGREEMENT_EXPIRED = 8;        /// The agreement end time has already passed.
-  REJECT_REASON_UNSUPPORTED_METADATA_VERSION = 9;  /// The metadata version is not supported.
+  REJECT_REASON_UNSPECIFIED = 0;                   /// Rejected for a reason not covered below (the catch-all).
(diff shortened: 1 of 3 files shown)


crates/dips/proto/indexer.proto
```diff
@@ -46,12 +46,18 @@ enum ProposalResponse {
 enum RejectReason {
   REJECT_REASON_UNSPECIFIED = 0;              /// Default / not set (used for ACCEPT responses).
   REJECT_REASON_PRICE_TOO_LOW = 1;            /// The offered price is below the indexer's minimum.
-  REJECT_REASON_OTHER = 2;                    /// Any other reason (bad signature, etc.).
+  REJECT_REASON_OTHER = 2;                    /// Any other reason not covered by a more specific variant.
   REJECT_REASON_SIGNER_NOT_AUTHORISED = 3;    /// The proposal signer is not authorised on the escrow contract.
```
RembrandtK: Should authorised on the Recurring Agreement Manager, not the escrow contract?

crates/dips/proto/indexer.proto
```diff
@@ -46,12 +46,18 @@ enum ProposalResponse {
...
   REJECT_REASON_DEADLINE_EXPIRED = 4;         /// The proposal deadline has already passed.
   REJECT_REASON_UNSUPPORTED_NETWORK = 5;      /// The subgraph's network is not supported by this indexer.
   REJECT_REASON_SUBGRAPH_MANIFEST_UNAVAILABLE = 6; /// The subgraph manifest could not be fetched from IPFS.
   REJECT_REASON_UNEXPECTED_SERVICE_PROVIDER = 7;   /// The RCA service provider does not match this indexer.
   REJECT_REASON_AGREEMENT_EXPIRED = 8;        /// The agreement end time has already passed.
```
RembrandtK: Not needed as will as `REJECT_REASON_DEADLINE_EXPIRED`? Might have been picked up to correlate to `endsAt` rather than `deadline', but do not need a reason for `endsAt`, should never be more restrictive than `deadline'.

It might however be worth adding a reason for invalid agreement proposals (that is proposals with invalid values).

crates/dips/proto/indexer.proto
```diff
@@ -46,12 +46,18 @@ enum ProposalResponse {
...
   REJECT_REASON_UNSUPPORTED_METADATA_VERSION = 9;  /// The metadata version is not supported.
+  REJECT_REASON_INVALID_SIGNATURE = 10;       /// The proposal's signature failed to verify.
+  REJECT_REASON_SENDER_NOT_TRUSTED = 11;      /// The proposal sender is not a trusted agreement manager.
```
RembrandtK: How does this differ from `REJECT_REASON_SIGNER_NOT_AUTHORISED`?

crates/dips/proto/indexer.proto
```diff
@@ -46,12 +46,18 @@ enum ProposalResponse {
...
+  REJECT_REASON_CAPACITY_EXCEEDED = 12;       /// The indexer has reached its DIPs agreement capacity.
+  REJECT_REASON_MANIFEST_TOO_LARGE = 13;      /// The subgraph manifest exceeds the maximum allowed size.
+  REJECT_REASON_REPLAY_DETECTED = 14;         /// The proposal repeats an already-seen message.
```
RembrandtK: Should think if idepotency is better and practical, and/or if resending is valid anyway (if a previous rejection was for capacity for example).

For the first Dipper could fail to get a response, try to resend, and get a failure response. In the meantime the indexer could have accepted?

crates/dips/proto/indexer.proto
```diff
@@ -46,12 +46,18 @@ enum ProposalResponse {
 enum RejectReason {
   REJECT_REASON_UNSPECIFIED = 0;              /// Default / not set (used for ACCEPT responses).
```
RembrandtK: Is this actually 'ACCEPTED' then, rather than an unspecified rejection? Or just do not define an enum member for 0? Could also logically overlap with `REJECT_REASON_OTHER`.

crates/dips/proto/indexer.proto
```diff
@@ -46,12 +46,18 @@ enum ProposalResponse {
...
   REJECT_REASON_PRICE_TOO_LOW = 1;            /// The offered price is below the indexer's minimum.
-  REJECT_REASON_OTHER = 2;                    /// Any other reason (bad signature, etc.).
+  REJECT_REASON_OTHER = 2;                    /// Any other reason not covered by a more specific variant.
```
RembrandtK: Might be worth allowing an optional text field for extra information (regardless of which reason)?

RembrandtK (commented):
I have reviewed at leaf level, a good set of options considered.

I wonder about the model of seeking to predict, how to group handling, and how to extend or provide more detail.

Might want to consider a model that splits and groups more by area than at leaf level of individual aspects. Having a MECE set of areas could help, for example the reason could be related to:
- Agreement Timing
- Agreement Conditions
- Agreement Price
- Agreement Metadata
- Agreement Parties
- Agreement Auth
- Escrow
- Subgraph

Might be able to orthogonally group into a set of what can be wrong, under two broad (extendable) categories:
- Invalid
	- Failed to retrieve or not available
	- Unable to parse
	- Incorrect parameters
	- Expired or no longer valid
- Declined
	- Not accepted (terms/price not accepted, party not accepted, etc)
	- Not supported (do not have capability or do not want to perform this type of work)
	- Too large/big (permanent capacity rejection)
	- Busy (could be be resubmitted later)

Additionally allowing optional extra text detail could help.

It might be not all combinations are applicable, but many would be. An extendable design would allow options to be added without breaking backwards compatibility, so that older software on either side could continue to work with still correct but less granular information.

crates/dips/proto/indexer.proto
MoonBoi9001: Done in ca26d2b

crates/dips/proto/indexer.proto
MoonBoi9001: Interesting point, I wanted to guard against a replay attack but that can be solved with other levers in a follow up PR.

RembrandtK (approved):

PR #1045, opened 2026-06-08
feat: verify the signature on indexing agreement proposals

MoonBoi9001:
## TL;DR

The indexer stopped checking the signature on incoming indexing-agreement proposals when authorization moved to a later on-chain step. This recovers the proposal's signer from its signature and turns away anything unsigned or tampered before the indexer acts on it. The recovered signer is the handle a following change will use to check the sender is one the indexer trusts.

## Motivation

When the dipper sends an indexing-agreement proposal to an indexer over gRPC, the indexer used to verify a signature on it; that check was dropped when authorization moved to a later on-chain step. As a result the indexer now fetches the subgraph manifest, prices it, and stores the proposal before it has any assurance the message wasn't forged or altered in flight, and without knowing who actually sent it.

This recovers the signer's address from the proposal's signature the moment it arrives and rejects anything missing or malformed, so a tampered or spoofed proposal is turned away before any work is done. On its own it authenticates the message; a following change will check that the recovered signer is one the indexer trusts to send proposals.

## Summary

- Recover the proposal signer from its signature over the recurring-collector signing domain.
- Reject proposals with a missing or malformed signature before any other validation.
- Add a recurring-collector address setting, required only when direct indexer payments are on.
- Fail startup if that address is missing while payments are enabled, instead of misverifying.
- Cover signer recovery, empty and malformed signatures, and the sign/recover round-trip with tests.

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -61,6 +61,7 @@ use thegraph_core::alloy::{
 
 #[cfg(feature = "db")]
 pub mod database;
+pub mod eip5267;
 pub mod inflight;
 pub mod ipfs;
 pub mod price;
@@ -74,6 +75,8 @@ pub mod store;
 
 use thiserror::Error;
 use uuid::Uuid;
+
+pub use crate::eip5267::fetch_rca_eip712_domain;
 
 /// Protocol version (seconds-based RCA)
 pub const PROTOCOL_VERSION: u64 = 2;
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -53,10 +53,10 @@ use std::sync::Arc;
 use server::DipsServerContext;
 use thegraph_core::alloy::{
     core::primitives::Address,
-    primitives::{keccak256, ruint::aliases::U256, Uint},
+    primitives::{keccak256, ruint::aliases::U256, Signature, Uint},
     signers::SignerSync,
     sol,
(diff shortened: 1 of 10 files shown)


crates/dips/src/lib.rs
```diff
@@ -147,9 +147,23 @@ fn derive_agreement_id(rca: &RecurringCollectionAgreement) -> Uuid {
     Uuid::from_bytes(id_bytes)
 }
 
+/// EIP-712 domain for RecurringCollectionAgreement signatures. Must match the
+/// domain dipper signs with and the on-chain RecurringCollector contract; any
+/// drift in name, version, chain id, or address makes every recovered signer wrong.
+pub fn rca_eip712_domain(chain_id: u64, recurring_collector: Address) -> Eip712Domain {
+    eip712_domain! {
```
RembrandtK, line 157: The contract uses OZ EIP712Upgradeable which implements EIP-5267 and exposes `eip712Domain()` returning the canonical (name, version, chainId, verifyingContract, …). Could we pull from that instead? One call covers everything and stays correct across redeploys/upgrades.

RembrandtK (changes requested):
I suggest calling `eip712Domain()` rather than hard-coding EIP 712 parameters. These are set by configuration at deployment time and could easily change, and from indexers perspective they would want validation the signature matches what is on chain.

RembrandtK (approved):

PR #1048, opened 2026-06-09
feat: reject proposals from untrusted senders and update docs

MoonBoi9001:
## TL;DR

The indexer's payments service accepts any agreement proposal with a cryptographically valid signature, without checking who sent it. This change requires the recovered signer to be an approved agreement manager on-chain — read from the indexing-payments subgraph — and rejects every other sender. It also carries a set of unrelated configuration-documentation fixes, called out separately below.

## Motivation

The payments service recovers the signer of each incoming agreement proposal but then accepts any valid signature, with no check on whether that signer is allowed to offer agreements. An earlier allowlist was removed because it checked a payer field that the caller writes into the proposal themselves, which anyone could spoof; the only identity the indexer can trust is the cryptographically recovered signer. Until an authorization step lands, any party that can reach the payments port can have a proposal accepted and stored, which is the wrong default to carry into a real deployment. This change reads the set of approved agreement managers from an on-chain source — the indexing-payments subgraph, which indexes the agreement-manager role assignments — and accepts a proposal only when its recovered signer is in that set.

## Summary

Authorize agreement senders (the main change):
- Authorize the recovered signer before any other work: it must hold the agreement-manager role.
- Read the approved set from the indexing-payments subgraph, cached and refreshed on a timer.
- Refresh on demand for an unknown signer (rate-limited), so a new manager is recognised in seconds.
- A signer rejected in the last hour is cached, so a later approval can take an hour to apply.
- Serve known senders from cache through a subgraph outage, within a bounded window, then reject.
- Reject an unknown sender clearly; report an unverifiable set as transient so the sender retries.
- Add config for the subgraph endpoint and the refresh, fail-open-grace, and staleness timings.
- Validate the DIPs collector and subgraph config at load, so misconfiguration fails fast.
- Emit metrics for proposal outcomes and role-set health so operators can see the gate working.

Unrelated configuration and documentation changes also in this PR:
- Fix contradictory notes in the example config and document the DIPs tunables.
- Clarify when to set the subgraph auth token and how local indexing relates to the query URL.
- Spell out the credit risk of trusted senders, and that the free-query token bypasses payment.

### DIFF
--- a/crates/config/maximal-config-example.toml
+++ b/crates/config/maximal-config-example.toml
@@ -194,6 +194,10 @@ supported_networks = []
 # Your own observations may differ - adjust pricing accordingly.
 min_grt_per_billion_entities_per_30_days = "200"  # entity-based component (global)
 
+# RecurringCollector contract address. The EIP-712 verifying contract used to
+# recover the signer of incoming agreement proposals. Required when DIPs is enabled.
+recurring_collector = "0xcccccccccccccccccccccccccccccccccccccccc"
+
 [dips.min_grt_per_30_days]  # base rate component (per-network)
 # arbitrum-one = "450"
 # matic = "300"
--- a/crates/config/maximal-config-example.toml
+++ b/crates/config/maximal-config-example.toml
@@ -94,9 +94,10 @@ max_signers_per_payer = 0
 # Indexing-payments subgraph, read for the set of accounts allowed to send DIPs
 # agreement proposals (the agreement-manager role). Required when [dips] is set.
 [subgraphs.indexing_payments]
-# Query URL for the indexing-payments subgraph.
+# Query URL for the indexing-payments subgraph -- a SEPARATE deployment from [subgraphs.network].
 query_url = "http://example.com/indexing-payments-subgraph"
-# How often to refresh contract information from this subgraph, in seconds.
+# Required when this section is present (no default). How often to refresh from this subgraph, in
+# seconds. Distinct from dips.role_refresh_interval_secs, which governs the role-set pull.
 syncing_interval_secs = 60
 # Optional, auth token used as "bearer auth".
 # query_auth_token = "super-secret"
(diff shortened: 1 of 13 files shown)


crates/config/maximal-config-example.toml
```diff
@@ -174,32 +194,55 @@ max_receipts_per_request = 10000
 0xDDE4cfFd3D9052A9cb618fC05a1Cd02be1f2F467 = "https://tap-aggregator.network.thegraph.com"
 0xDD6a6f76eb36B873C1C184e8b9b9e762FE216490 = "https://tap-aggregator-arbitrum-one.graphops.xyz"
 
-# DIPs (Direct Indexer Payments). Payer authorisation is verified
```
RembrandtK, line 177: Looking through code, often the correct acronym of 'Direct Indexing Payments' is not used?

crates/config/src/config.rs
```diff
@@ -682,6 +701,14 @@ pub struct DipsConfig {
     /// RecurringCollector address: the EIP-712 verifying contract used to recover
     /// the signer of incoming RCA proposals. Required when DIPs is enabled.
     pub recurring_collector: Option<Address>,
+    /// How often to re-pull the agreement-manager role set from the subgraph.
+    pub role_refresh_interval_secs: u64,
+    /// Extra time past the refresh interval that a cached role set stays trusted
+    /// while refreshes fail, before the gate fails closed (the fail-open window).
+    pub role_failopen_grace_secs: u64,
+    /// Max seconds the role subgraph head may lag wall-clock before its data is
+    /// treated as unreliable. 0 disables the check.
```
RembrandtK, line 710: 0 should mean what it says, no lag acceptted, not invert to turning it off?

crates/config/maximal-config-example.toml
```diff
@@ -91,6 +92,22 @@ escrow_min_balance_grt_wei = "100000000000000000"
 # signers, orphaning receipts and enabling free queries.
 max_signers_per_payer = 0
```
RembrandtK, line 93: Not part of this PR, I would suggest not using 0 to mean unlimited. (Can use no value, as done elsewhere.)

RembrandtK (approved):
Some minor observations made in comments. Looks good. :)

crates/config/maximal-config-example.toml
MoonBoi9001: Was addressed here, but currently only on main, needs bringing into `main-dips-rebased` https://github.com/graphprotocol/indexer-rs/pull/1040

crates/config/src/config.rs
MoonBoi9001: Going to defer this for a follow up, will link to it when ready

crates/config/maximal-config-example.toml
MoonBoi9001: Going to defer this for a follow up, will link to it when ready

crates/config/src/config.rs
MoonBoi9001: This stack targets this work:

- #1056
- #1057
- #1058 
- #1059

PR #1050, opened 2026-06-09
feat: introduce optional cap on # of DIPs agreements accepted per day

MoonBoi9001:
## TL;DR

Indexers can now set a limit on how many DIPs agreements they take on in a day. Once an indexer holds that many live agreements (awaiting acceptance or accepted) in the last 24 hours, further proposals are turned away with a capacity reason. The limit is off by default, so nothing changes unless an operator sets it. Intoduced by indexer demand.

## Motivation

When the dipper sends an indexer a DIPs proposal, the indexer validates it and stores it for its agent to act on, with no ceiling on how many it will take in a day. A burst of proposals can therefore commit an indexer to indexing far more subgraphs than it has capacity for, and today the only limits are transport-level rate limits that bound request volume, not the indexing work each stored proposal implies. This gives operators a domain-level safeguard: a per-indexer daily cap on agreements taken on, so an indexer can protect its own resources. The cap is best-effort and opt-in, not a hard guarantee.

## Summary

- Add an optional per-day cap on live DIPs agreements an indexer holds, over a rolling 24h window.
- Over the cap, reject the proposal with the existing capacity reason before any manifest download.
- Count live agreements (pending or accepted) from the database; exclude rejected and expired.
- Default to no cap; an unreadable count fails open, so a database blip doesn't reject proposals.
- Cover the behaviour with unit tests and database-backed tests for the window and status filter.

### DIFF
--- a/crates/config/maximal-config-example.toml
+++ b/crates/config/maximal-config-example.toml
@@ -204,7 +204,7 @@ recurring_collector = "0xcccccccccccccccccccccccccccccccccccccccc"
 
 # Optional safeguard: cap live DIPs agreements (awaiting acceptance or accepted) per
 # rolling 24h; rejected and expired proposals don't count. Unset = no cap; 0 rejects all.
-# max_agreements_per_day = 30
+# max_new_agreements_per_24h = 30
 
 [dips.min_grt_per_30_days]  # base rate component (per-network)
 # arbitrum-one = "450"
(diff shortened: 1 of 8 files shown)


crates/config/maximal-config-example.toml
```diff
@@ -202,6 +202,10 @@ recurring_collector = "0xcccccccccccccccccccccccccccccccccccccccc"
 # domain at startup (EIP-5267). When unset, built-in domain constants are used.
 # rpc_url = "https://arb1.arbitrum.io/rpc"
 
+# Optional safeguard: cap live DIPs agreements (awaiting acceptance or accepted) per
+# rolling 24h; rejected and expired proposals don't count. Unset = no cap; 0 rejects all.
+# max_agreements_per_day = 30
```
RembrandtK: Seems this is not per day, closer to `max_new_agreements_per_24h`?

RembrandtK (approved):

PR #1061, opened 2026-06-19
fix: store agreement proposals without the signer signature

MoonBoi9001:
## TL;DR

This makes the indexer service remove the signature before saving an indexing proposal.

## Motivation

When the service receives an indexing-agreement proposal it first verifies the sender's signature, then saves the proposal for the agent to act on. The agent no longer uses that signature to accept on-chain — it accepts against the payer's posted on-chain offer — so it now treats any saved proposal that still carries a signature as a mistake and refuses it. The result is that every proposal the service saves is discarded by the agent and no agreement can be accepted. Clearing the signature before saving, while still verifying it first, lets the two services line up again.

## Summary

- Keeps verifying the sender's signature on every incoming proposal, unchanged.
- Saves the proposal with the signature cleared, which is the form the agent expects.
- Compares a re-sent proposal against that same cleared form so genuine retries still match.
- Updates the affected tests to seed and check the stored form.

Generated with [Claude Code](https://claude.com/claude-code)
https://claude.ai/code/session_01USBDULGGvrHzVTKWgxfYz4

### DIFF
--- a/crates/dips/src/lib.rs
+++ b/crates/dips/src/lib.rs
@@ -276,6 +276,17 @@ impl SignedRecurringCollectionAgreement {
     pub fn encode_vec(&self) -> Vec<u8> {
         self.abi_encode()
     }
+
+    /// ABI-encode the same agreement but with an empty signature. The signature
+    /// only authenticates the sender at proposal time; the indexer-agent rejects
+    /// any stored proposal that still carries one (acceptance is offer-based).
+    pub fn encode_without_signature(&self) -> Vec<u8> {
+        SignedRecurringCollectionAgreement {
+            agreement: self.agreement.clone(),
+            signature: Default::default(),
+        }
+        .abi_encode()
+    }
 }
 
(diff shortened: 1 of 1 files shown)


tmigone:
what about doing the opposite? i.e, having the service store the signature but the agent to ignore it. Is there any downside to that? There isn't an immediate use for the signature, but it is proof that can be potentially valuable.

MoonBoi9001:
Makes sense, will look into it

MoonBoi9001:
Closing as superseded. Store-with-signature — the suggestion in review here — already shipped on `main-dips-rebased` via the #1042–#1052 series (#1045 verifies the signature, and the service stores the raw signed bytes). This PR's only remaining unique change was stripping the signature before storage, i.e. the opposite, so there's nothing left to land.

The other half — having the indexer-agent ignore the stored signature instead of rejecting it as a `non_empty_signature` regression — will land separately in graphprotocol/indexer, stacked on graphprotocol/indexer#1240.

PR #1069, opened 2026-07-10
perf: stop querying the network subgraph on every TAP receipt

MoonBoi9001:
This PR makes the allocation-redeemed receipt check read the indexer's own database instead of issuing a live network-subgraph query per receipt. That lookup sat in the hot path of every paid query: an external round trip (billed, for indexers whose network subgraph is gateway-served) spent asking a question whose answer almost never matters in the routine lifecycle. tap-agent only writes the closing RAV when an allocation ages out of the recently-closed buffer, indexer-agent redeems only closing RAVs, and the eligibility check stops accepting receipts at that same buffer expiry, so by the time a redemption exists, receipts for that allocation are already rejected 1 check earlier. The redeemed check is really insurance against out-of-band redemptions and agent-service config skew, and insurance should not cost a network call per receipt.

The replacement keeps exactly that insurance at zero hot-path cost. indexer-agent already records redemptions in `redeemed_at` (and reconciles them against the chain in both directions, including clearing on reorg), so the check now holds an in-memory set of redeemed closing RAVs, loaded at startup and updated through a new trigger with pg_notify, the deny-list check's pattern. Rows are scoped to this service provider and keyed by payer, data service, and collection, matching the receipt's own fields.


tmigone:
I've looked at the PR, even if the underlying change is good I'd like to discuss the possibility of making bigger changes here. This PR persists the legacy/v1 invariant of 1 RAV per allo. It might be time for us to modernize the stack for horizon.

MoonBoi9001:
@tmigone Sounds good to me, I can explore a query payment decoupling refactor

tmigone (approved):
lgtm.

I was a bit concerned about a check depending entirely on state written by indexer agent, but I think it's acceptable. If the agent stops, this check silently no-ops and `AllocationEligible` still covers the routine case. The tradeoff is that it moves the source of truth from chain state to agent bookkeeping, so it's now best-effort insurance rather than authoritative. I think fine for what this check is.

PR #1078, opened 2026-08-05
fix: reject DIPs proposals whose subgraph files are missing

MoonBoi9001:
This PR makes DIPs proposal validation verify that every file a subgraph manifest links (schema, mapping WASM, ABIs) is retrievable from IPFS before accepting, and rejects with the existing SUBGRAPH_MANIFEST_UNAVAILABLE reason when any is not. Validation previously fetched only the manifest, which does not prove the subgraph is deployable: graph-node resolves all linked files at deploy time, so a subgraph with one missing file can never be deployed, while the agreement paying for it is already accepted.

### DIFF
--- a/crates/dips/src/ipfs.rs
+++ b/crates/dips/src/ipfs.rs
@@ -1,7 +1,7 @@
 // Copyright 2023-, Edge & Node, GraphOps, and Semiotic Labs.
 // SPDX-License-Identifier: Apache-2.0
 
-//! IPFS client for fetching subgraph manifests.
+//! IPFS client for fetching subgraph manifests and verifying the files they link.
 //!
 //! When validating an RCA, we need to verify that the referenced subgraph
 //! deployment actually exists and determine which network it indexes.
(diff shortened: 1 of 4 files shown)


crates/dips/src/ipfs.rs
```diff
@@ -199,37 +246,145 @@ impl IpfsClient {
                 ))
             })?
     }
+
+    /// Stream a linked file to prove it is retrievable, counting bytes against
+    /// the cap and discarding them so peak memory is one chunk, not the file.
```
tmigone, line 251: We discard it because there is no use for the indexer-rs besides proving it can be retrieved no? 
Bit sad that graph node then has to re-fetch the same files but given the size and complexity i dont think it's worth trying to streamline this.

tmigone (approved):
lgtm, only comment is to consider wether differentiating between the manifest being unavailable and the linked files being unavailable is useful or not. I suspect it's not, but if it is we might want to show different errors/failure modes.

crates/dips/src/ipfs.rs
MoonBoi9001: >Bit sad that graph node then has to re-fetch the same files

Was also thinking this, maybe something to revisit at another point in time.

MoonBoi9001: >only comment is to consider wether differentiating between the manifest being unavailable and the linked files being unavailable is useful or not

Yeah I was leaning away from this as then dipper would need handling on another rejection reason, and thought in this instance its easier to bundle them into one to reuse the same handling on dipper's side. Would not be too hard to disintangle them in the future, just did not see a huge reason to be concerned about it at this time.

PR #1079, opened 2026-08-05
fix: answer DIPs proposals inside default proxy timeouts

MoonBoi9001:
This PR bounds all IPFS work in DIPs proposal validation to a 40-second budget, with a 10-second floor reserved for verifying the files the manifest links, so validation always answers within 50 seconds. Validation fetches a subgraph's manifest and its linked files from IPFS before accepting, retrying with backoff when IPFS is slow; the worst case was 220 seconds of silence on the gRPC stream, and most reverse proxies kill a quiet stream at 60 seconds by default (nginx included). On the slow path the payer therefore saw a dead call instead of a rejection and its reason, and every indexer behind a default proxy config hits the same wall. A tighter budget means slow-but-alive IPFS content can be rejected.

### DIFF
--- a/crates/dips/src/ipfs.rs
+++ b/crates/dips/src/ipfs.rs
@@ -34,10 +34,11 @@
 //!
 //! # Timeout, Retry, and Size Limits
 //!
-//! IPFS fetches have a 30-second timeout per attempt and retry up to 3 times
-//! with exponential backoff (5s, 10s, 20s delays), but the whole phase is
-//! bounded by `IPFS_PHASE_BUDGET`: 40 seconds for the manifest and its linked
-//! files together, so validation answers inside the 60-second gRPC stream
+//! IPFS fetches get 2 attempts of 15 seconds each with a 2-second pause
+//! between them; a fetch that succeeds at all almost always does so well
+//! inside 15 seconds. Both attempts and the pause fit inside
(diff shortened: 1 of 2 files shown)


crates/dips/src/ipfs.rs
```diff
@@ -89,16 +92,20 @@ use crate::{
...
 pub(crate) const IPFS_FETCH_TIMEOUT: Duration = Duration::from_secs(30);
 
-/// Ceiling on all IPFS work for one proposal, manifest and linked files together;
-/// equals the manifest fetch's own retrying worst case (30+10+30+20+30+40+30s),
-/// so adding linked-file verification did not grow the documented budget.
-pub(crate) const IPFS_PHASE_BUDGET: Duration = Duration::from_secs(190);
+/// Ceiling on all IPFS work for one proposal, manifest and linked files together,
+/// sized so validation answers inside the 60-second gRPC stream timeout most
+/// reverse proxies default to, which would turn a rejection into a dead call.
+pub(crate) const IPFS_PHASE_BUDGET: Duration = Duration::from_secs(40);
```
tmigone, line 101: Worth noting that with a 30s timeout this means IPFS work is mostly a single attempt (no retries). That is probably ok,

tmigone (approved):

PR #1080, opened 2026-08-07
fix: update rand sample_iter usage for rand 0.10 API

ayushsingh82:
## Summary

The `rand` crate v0.10 moved `sample_iter` off the `Rng` trait and onto `Distribution` — it's now called as `distribution.sample_iter(rng)` instead of `rng.sample_iter(&distribution)`. This currently breaks `cargo clippy` and `cargo test` on the pending `rand` 0.10.0 security-update PR (#1070), in `indexer-tap-agent`'s `test.rs` (confirmed via CI logs on #1070) and `indexer-dips`'s `lib.rs` test module (same broken pattern, not exercised by that CI job but would break on `main` once #1070 merges).

This PR applies the same fix pattern used in `rand`'s own docs (calling `sample_iter` on the `Distribution` instance) in both places, and drops the now-unused `Rng` import.

## Changes
- `crates/tap-agent/src/test.rs`: `generate_random_prefix()` fixed
- `crates/dips/src/lib.rs`: `VoucherContext::random()` (test-only) fixed

## Test plan
- [x] `cargo check -p indexer-tap-agent --tests` — compiles clean
- [x] `cargo check -p indexer-dips --tests` — compiles clean
- [x] `cargo fmt --check -p indexer-tap-agent -p indexer-dips` — no diff
- [x] `cargo clippy -p indexer-tap-agent -p indexer-dips --all-features --all-targets -- -A dead-code -D warnings` — clean, matching CI's exact invocation

This should unblock #1070 (`rand` 0.10.0 [security]) once rebased on top of, or merged alongside, this fix.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1895,7 +1895,7 @@ version = "3.9.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6dee98b0db6a962de883bf5d20362dee4d7ca0d12fe39a7c6c73c844e1cd7c1f"
 dependencies = [
- "darling 0.20.11",
+ "darling 0.23.0",
  "ident_case",
  "prettyplease",
  "proc-macro2",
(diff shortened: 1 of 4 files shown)


ayushsingh82:
cc @MoonBoi9001 — saw you're active on tap-agent/dips lately. Quick heads up: rand 0.10 moved `sample_iter` from `Rng` to `Distribution`, which is currently breaking clippy/tests on the pending rand 0.10.0 security bump (#1070). This PR fixes the two broken call sites in tap-agent's test.rs and dips' lib.rs so that bump can land cleanly.

PR #1081, opened 2026-08-07
fix: update prometheus label API usage for prometheus 0.14

ayushsingh82:
## Summary

`prometheus` 0.14 tightened `with_label_values` / `get_metric_with_label_values` / `remove_label_values` to require a single uniform label type across the slice. Several call sites mixed `&String` (from `.to_string()`) with `&str` constants (`TAP_VERSION` / `TAP_V2`) in the same array, which no longer type-checks. This currently breaks CI on the pending `prometheus` 0.14.0 bump (#1027).

Also fixes one `deprecated` warning (`LabelPair::get_value()` → `.value()`) that would otherwise fail the workspace's `cargo clippy -D warnings` gate once the above compiles.

## Changes
- `crates/tap-agent/src/agent/sender_allocation.rs`: 4 call sites normalized to `&str` via `.as_str()`
- `crates/tap-agent/src/agent/sender_account.rs`: 7 call sites (incl. 3 in tests) normalized the same way
- `crates/service/src/middleware/prometheus_metrics.rs`: replaced deprecated `get_value()` with `value()`

## Test plan
- [x] `cargo check --workspace --all-targets --all-features` — compiles clean
- [x] `cargo clippy --workspace --all-features --all-targets -- -A dead-code -D warnings` — clean, matching CI's exact invocation
- [x] `cargo fmt --check` — no diff

Should unblock #1027 (`prometheus` 0.14.0) once rebased on top of, or merged alongside, this fix.


ayushsingh82:
cc @MoonBoi9001 — same theme as #1080: this one fixes 11 `with_label_values`/`get_metric_with_label_values`/`remove_label_values` call sites in tap-agent (sender_allocation.rs, sender_account.rs) that mixed `&String` and `&str` labels, which prometheus 0.14 no longer allows. Also fixed one deprecated-API warning in indexer-service-rs that would've failed the clippy -D warnings gate. Unblocks the pending prometheus 0.14.0 bump (#1027).

PR #1086, opened 2026-08-11
fix(tap-agent): accept receipts newer than the last redemption

Maikol:


crates/tap-agent/src/tap/context/checks/allocation_id.rs
```diff
@@ -79,14 +86,26 @@ impl Check<TapReceipt> for AllocationId {
...
+        let last_redeemed_at_ns = last_redeemed_at_secs
+            .checked_mul(NANOS_PER_SECOND)
+            .ok_or_else(|| {
+                CheckError::Failed(anyhow!(
+                    "Last redeemed timestamp {last_redeemed_at_secs}s overflows when converted to nanoseconds"
+                ))
+            })?;
+        let receipt_timestamp_ns = receipt.signed_receipt().timestamp_ns();
+
+        if receipt_timestamp_ns <= last_redeemed_at_ns {
```
MoonBoi9001, line 101: Here we're comparing the `receipt_timestamp_ns` to `last_redeemed_at_ns`, which comes from the block timestamp of the redeem transaction. When tap-agent collects, it excludes any receipts newer than `timestamp_buffer_secs`. So I think there's a gap here where some receipts will be lost.

crates/tap-agent/src/tap/context/checks/allocation_id.rs
```diff
@@ -128,18 +148,44 @@ async fn query_network_redeem_transactions(
         )
         .await?;
 
-    Ok(!data.payments_escrow_transactions.is_empty())
+    let mut latest_redeemed_at_secs: Option<u64> = None;
```
MoonBoi9001, line 151: With no `first`, `orderBy` or `orderDirection`, graph-node will return 100 rows ordered by id.

Adding `first: 1` would break `sender_account.rs`. So this check could use it's own query:

```
query LatestRedeemTransactionQuery($payer: Bytes!, $receiver: Bytes!, $allocationId: Bytes!) {
  paymentsEscrowTransactions(
    first: 1
    orderBy: timestamp
    orderDirection: desc
    where: { type: "redeem", payer_: { id: $payer }, receiver_: { id: $receiver }, allocationId: $allocationId }
  ) { timestamp }
}
```

